PR #107, opened 2018-11-01
Datastore refactor

paulwhomer:
Refactored datastore, cleaned up error handling and merged lots of branches.


node/app/query.go
```diff
@@ -196,63 +163,38 @@ func HandleUtxoQuery(app Application, message []byte) []byte {
...
-
-func HandleVersionQuery(app Application, message []byte) []byte {
-	return []byte(version.Current.String())
+func HandleVersionQuery(app Application, message []byte) interface{} {
+	return version.Current.String()
 }
 
 // Get the account information for a given user
-func HandleBalanceQuery(app Application, message []byte) []byte {
+func HandleBalanceQuery(app Application, message []byte) interface{} {
```
alexjlan: Why change all []byte to interface{}? Isn't []byte more reliable?

paulwhomer: Interface preserves the Go types. Byte arrays flatten the data (through techniques like serialization). The handlers now pass up their native responses, it is flattened once at the top (so you no longer have to worry about serialization or hex or anything in the actual queries).

PR #111, opened 2018-11-05
Add some rpcs

reblws:


node/sdk/pb/base.proto
```diff
@@ -0,0 +1,81 @@
...
+
+message CheckAccountReply {
+    string name = 1;
+    string chain = 2;
+    bytes accountKey = 3;
+    // TODO: Should use Coin message
+    Balance balance = 4;
+}
+
+enum Currency {
```
alexjlan: Can you make this come from the /node/data/coin.go? so when we add new currency, we don't need to change both place?

node/sdk/pb/base.proto
```diff
@@ -0,0 +1,81 @@
...
+
+message SwapRequest {}
+message SwapReply {}
+
+message RegisterRequest {
+    string identity = 1;
+    ChainType chain = 2;
+}
+
+enum ChainType {
```
alexjlan: This should be align with node/data/chain.go as well.

PR #113, opened 2018-11-06
Signature

sergei-sachkov:
I introduced a method that creates a signature for an OLT transaction and cleaned up a Parse method as a preparation for publishing a transaction with its signature to a blockchain


node/action/base.go
```diff
@@ -58,6 +58,11 @@ type Transaction interface {
 	Resolve(interface{}, Commands)
 }
 
+type SignedTransaction struct {
+	Transaction
+	signature []byte
```
reblws: If you need to access this field outside the `action` package you need to capitalize it.

sergei-sachkov: @reblws , sorry, what should I do?

sergei-sachkov: Do you mean, I need to use Signature name?

paulwhomer: Yes, in Go lowercase means private, while Uppercase means public.

node/app/query.go
```diff
@@ -66,6 +70,51 @@ func HandleNodeNameQuery(app Application, message []byte) interface{} {
 	return global.Current.NodeName
 }
 
+func HandleSignTransaction(app Application, message []byte) interface{} {
+	var tx action.Transaction
+
+	transaction, transactionErr := serial.Deserialize(message, tx, serial.CLIENT)
+
+	signature := []byte{}
```
paulwhomer: So, this is returning the signed transaction to the client? That's okay, but I think we discussed having the fullnode itself do the signing and broadcast? I may have been wrong about that. If the fullnode does it, it will have to be async or a goroutine (so that it doesn't take too long and mess up consensus). But there isn't a security issue if you do send it back to the client and then it's the client that waits for the transaction to get committed. That offers some degree of error handling, at the expensive of making the client actions take longer...

node/app/application.go
```diff
@@ -274,9 +274,17 @@ func (app Application) DeliverTx(tx []byte) ResponseDeliverTx {
...
 	if result.ShouldProcess(app) {
-		ttype, _ := action.UnpackMessage(action.Message(tx))
+		transaction, _ := action.Parse(action.Message(tx))
+
+		// TODO: can we improve this?
+		swapProcessing := false
+		switch transaction.(type) {
+		case *action.Swap:    swapProcessing = true
+		case *action.Publish: swapProcessing = true
+		case *action.Verify:  swapProcessing = true
```
paulwhomer: golang will let you put all 3 in the same case statement, separated by commas...

PR #145, opened 2018-11-22
Olfullnode config

paulwhomer:
Changed the olfullnode to load its configuration from the config files. Tendermint still needs individual env variables, so there is now an olconfig tool to read the files using viper and get them into the shell.


node/cmd/olconfig/get.go
```diff
@@ -0,0 +1,115 @@
...
+	RootCmd.AddCommand(getCmd)
+
+	// TODO: I want to have a default account?
+	// Transaction Parameters
+	getCmd.Flags().StringArrayP("parameter", "p", arguments.names, "account name")
+	getCmd.Flags().Bool("peers", false, "handle peers")
+
+}
+
+// IssueRequest sends out a sendTx to all of the nodes in the chain
```
reblws: What is IssueRequest

PR #154, opened 2018-11-29
Added validators at the end of a block

sergei-sachkov:


alexjlan:
@sergei-sachkov The code looks okay, did you test it?

PR #162, opened 2018-12-06
Transaction validations

sergei-sachkov:
Strengthened validation rules for all transaction types

### DIFF
--- a/node/action/register.go
+++ b/node/action/register.go
@@ -31,6 +31,12 @@ func init() {
 func (transaction Register) Validate() status.Code {
 	log.Debug("Validating Register Transaction")
 
+	baseValidate := transaction.Base.Validate()
+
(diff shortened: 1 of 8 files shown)


node/action/register.go
```diff
@@ -31,13 +31,34 @@ func init() {
...
+		log.Debug("Missing AccountKey", "transaction", transaction)
+		return status.MISSING_DATA
+	}
+
+	if transaction.TendermintAddress == "" {
+		log.Debug("Missing TendermintAddress", "transaction", transaction)
+		return status.MISSING_DATA
+	}
+
+	if transaction.TendermintAddress == "" {
```
alexjlan: why we have 	if transaction.TendermintAddress == "" {
 two times? it probably be TendermintPubKey?

PR #165, opened 2018-12-11
Olvm client libs

steleeoneledger:
- The Javascript library is separated from go code source. 
- "olvm test" can run the smart contract code off the blockchain


.gitignore
```diff
@@ -1,7 +1,9 @@
 vendor
+/node/.vendor-new/*
```
paulwhomer: I think this is from an inconsistent state in dep, so it shouldn't be in the .gitignore...

node/Gopkg.lock
```diff
@@ -1,25 +1,6 @@
 # This file is autogenerated, do not edit; changes may be undone by the next 'dep ensure'.
 
 
-[[projects]]
```
paulwhomer: Shouldn't check in this file, unless we've explicitly changed the toml file to update the dependencies. We're not quite doing this correctly right now, but most time I just roll back these changes from the update, before committing (since they may cause problems with other developers setups).

PR #196, opened 2018-12-19
transfer amount arg VT to Payment account for NewValidators

pat-oneledger:


pat-oneledger:
- fixed olclient list --validators: if empty, do not print validators; fixed []uint8, not string problem
- fixed testmint of VTs: fee is fixed to OLT
- fixed olclient applyValidator: 
    * pass errors to show on console
    * if normal node, amount in the args is deducted from node's VT balance and transferred to Payment account
    * if normal node, can purge itself no VT needed
    * if administrator node (has more than 10 VTs), you can run applyValidator for another identity. That identity has to have at least 1 VT.
    * if administrator node (has more than 10 VTs), you can purge another validator.

PR #235, opened 2019-03-13
add the externalp2paddress for tendermint

alexjlan:


node/scripts/data/Bob.toml
```diff
@@ -2,6 +2,7 @@ Identity = "Bob"
 NodeName = "Bob-Node"
 RpcAddress = "tcp://127.0.0.1:26603"
 P2PAddress = "tcp://127.0.0.1:26613"
+ExternalP2PAddress = "tcp://127.0.0.1:26614"
```
reblws: This is the same as Carol's

PR #244, opened 2019-03-28
Fix addrbook not saving

reblws:


node/config/server.go
```diff
@@ -30,6 +30,50 @@ type Server struct {
...
+	baseConfig.DBBackend = leveldb
+	baseConfig.DBPath = "data"
+	baseConfig.LogLevel = cfg.Consensus.LogLevel
+
+	p2pConfig := cfg.P2P.TMConfig()
+	p2pConfig.ListenAddress = cfg.Network.P2PAddress
+	if cfg.Network.ExternalP2PAddress == "" {
+		p2pConfig.ExternalAddress = cfg.Network.P2PAddress
+	} else {
+		p2pConfig.ExternalAddress = cfg.Network.ExternalP2PAddress
```
harshad1ledger: again can avoid else block here if externalP2P address is assigned before if

PR #246, opened 2019-04-03
Serialize

harshad1ledger:


node/comm/client.go
```diff
@@ -198,8 +198,21 @@ func Query(path string, packet []byte) interface{} {
 		return nil
 	}
 
-	var prototype interface{}
-	result, err := serial.Deserialize(response.Response.Value, prototype, serial.CLIENT)
+	var result interface{}
+
+	if path == "/applyValidators" ||
```
reblws: The paths and handlers should probably be stored in a map so you can do something like:

```golang
	var path string
	paths := map[string]func(app.Application, map[string]interface{}) interface{}{
		"/applyValidators":     app.HandleApplyValidatorQuery,
		"/createExSendRequest": app.HandleCreateExSendRequest,
		"/createSendRequest":   app.HandleCreateSendRequest,
		"/createMintRequest":   app.HandleCreateMintRequest,
		"/createSwapRequest":   app.HandleSwapRequest,
		"/nodeName":            app.HandleNodeNameQuery,
		"/identity":            app.HandleIdentityQuery,
		"/signTransaction":     app.HandleSignTransaction,
		"/accountPublicKey":    app.HandleAccountPublicKeyQuery,
		"/accountKey":          app.HandleAccountKeyQuery,
		"/account":             app.HandleAccountQuery,
		"/validator":           app.HandleValidatorQuery,
		"/balance":             app.HandleBalanceQuery,
		"/version":             app.HandleVersionQuery,
		"/currencyAddress":     app.HandleCurrencyAddressQuery,
		"/sequenceNumber":      app.HandleSequenceNumberQuery,
		"/testScript":          app.HandleTestScript,
	}

	_, ok := paths[path]
	if ok {
		var proto interface{}
		result, err := serial.Deserialize(response.Response.Value, proto, serial.CLIENT)
	} else {
		err := clSerializer.Deserialize(response.Response.Value, &result)
	}
```

node/comm/client.go
```diff
@@ -198,8 +198,21 @@ func Query(path string, packet []byte) interface{} {
...
+		path == "/createExSendRequest"||
+		path == "/createSendRequest" ||
+		path == "/createMintRequest" ||
+		path == "/createSwapRequest" ||
+		path == "/nodeName" ||
+		path == "/signTransaction" {
+
+		var proto interface{}
```
reblws: Add comment why we use old serializer

node/id/identity.go
```diff
@@ -40,6 +41,7 @@ type Identity struct {
 
 func init() {
 	serial.Register(Identity{})
+	serialize.RegisterConcrete(new(Identity), "id_identity")
```
reblws: I think we should pass string constants for to `RegisterConcrete`

node/serialize/amino_strategy.go
```diff
@@ -0,0 +1,58 @@
...
+	bz, err := a.codec.MarshalBinaryLengthPrefixed(obj)
+
+	return bz, err
+}
+
+//Deserialize
+func (a *aminoStrategy) Deserialize(src []byte, dest interface{}) error {
+
+	/*
+	if apr, ok := dest.(adapter.DataAdapter); ok {
```
reblws: Remove commented code

PR #263, opened 2019-05-14
Register self on setupState

reblws:


app/application.go
```diff
@@ -211,11 +243,12 @@ type closer interface {
 	Close()
 }
 
-func newContext(cfg config.Server, logWriter io.Writer) (context, error) {
+func newContext(logWriter io.Writer, cfg config.Server, nodeCtx *NodeContext) (context, error) {
```
harshad1ledger: just a question, any thought behind changing the order?

PR #279, opened 2019-05-27
OLP 284

harshad1ledger:


data/balance/coin.go
```diff
@@ -114,24 +116,27 @@ func (coin Coin) Equals(value Coin) bool {
...
 		debug.PrintStack()
 		logger.Fatal("Invalid Coin", coin)
 	}
 
 	if coin.Currency.Name != value.Currency.Name {
 		//logger.Error("Mismatching currencies", "coin", coin, "value", value)
-		logger.Fatal("Mismatching currencies", coin, value)
-		return coin
+		logger.Error("Mismatching currencies", coin, value)
+		return coin, errors.New("")
```
alexjlan: error info empty

data/balance/coin.go
```diff
@@ -114,24 +116,27 @@ func (coin Coin) Equals(value Coin) bool {
 }
 
 // Minus two coins
-func (coin Coin) Minus(value Coin) Coin {
+func (coin Coin) Minus(value Coin) (Coin, error) {
```
alexjlan: if you change the Minus return type, probably also change the Plus, and so on

data/balance/coin.go
harshad1ledger: fixed this

data/balance/coin.go
harshad1ledger: fixed this

PR #300, opened 2019-06-20
added handlers for ons transactions

harshad1ledger:


client/request_types.go
```diff
@@ -118,3 +118,50 @@ type BroadcastTxCommitReply struct {
 	TxHash keys.Address                   `json:"txHash"`
 	Result ctypes.ResultBroadcastTxCommit `json:"result"`
 }
+
+/*
+	ONS Request Types
+*/
+type OnsCreateRequest struct {
```
reblws: Since ONS is an acronym I feel it should be `ONSCreateRequest`

client/request_types.go
```diff
@@ -118,3 +118,50 @@ type BroadcastTxCommitReply struct {
...
+	Owner   []byte        `json:"owner"`
```
reblws: When dealing with addresses use the `keys.Address` type so it marshals into the form `0x...` in JSON

reblws: Otherwise it gets encoded as base64 in the JSON

service/tx/ons.go
```diff
@@ -0,0 +1,159 @@
...
+	fee := action.Fee{args.Fee, args.Gas}
+	tx := &action.BaseTx{
+		Data: domainSend,
+		Fee:  fee,
+		Memo: uuidNew.String(),
+	}
+
+	packet, err := serialize.GetSerializer(serialize.NETWORK).Serialize(tx)
+	if err != nil {
+		return errors.Wrap(err, "err while network serialization")
```
reblws: When returning rpc errors use `rpc.NewError` or one of the `rpc.Code` related errors. I'd use `rpc.InternalError` in this case

reblws: Do this for all errors you return

client/request_types.go
reblws: Same with all other `Ons` prefixes

PR #323, opened 2019-07-18
add name for domain struct

Ruins7:


data/balance/amount.go
```diff
@@ -12,7 +12,7 @@ import (
 
 // Amount represents an amount of a currency
 type Amount struct {
-	Int big.Int `json:"bint"`
+	Int big.Int `json:"bint,string"`
```
reblws: I don't think `,string` does anything for `big.Int`

reblws: The JSON codec already encodes `big.Int `as `string`

PR #343, opened 2019-09-09
Olp 438

harshad1ledger:


service/codes/codes.go
```diff
@@ -0,0 +1,17 @@
+/*
+
+ */
+
+package codes
+
+const (
+	InvalidParams         = 1001
```
Ruins7: maybe change code number start from 3 and 5 digits long, because in SDK, request error code starts with 1, response error code start with 2 and protocol error code start from 3. we should make it consistent

PR #390, opened 2020-01-20
Erc20 lock

kingpinXD:
ERC20 Lock and redeem.


kingpinXD:
Had initialy taken a Branch from ons-store , thus merging back into it .
Will merge it to develop together with ons

PR #428, opened 2020-02-21
OLP-648: returning an error instead of panicking when invalid private…

KevinOneLedger:
… key is passed.

### DIFF
--- a/cmd/olfaucet/main.go
+++ b/cmd/olfaucet/main.go
@@ -323,6 +323,12 @@ func NewFaucet(cfg *config.Server) (*Faucet, error) {
 		return nil, err
 	}
 
+	_, err = nodeCtx.PrivKey().GetHandler()
+	if err != nil {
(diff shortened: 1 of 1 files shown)


cmd/olfaucet/main.go
```diff
@@ -259,12 +259,12 @@ func (f *Faucet) RequestOLT(req Request, reply *Reply) error {
 	rawTx := sendTxResults.RawTx
 	h, err := f.nodeCtx.PrivKey().GetHandler()
 	if err != nil {
-		panic("invalid nodeCtx private key")
+		return errors.New("invalid nodeCtx private key")
```
alexjlan: this will make the faucet running with errors, we should check the key is correct when launch the faucet.

KevinOneLedger: what is the suggestion for this? Should I just cancel the pull request and leave it as a panic instead?

PR #430, opened 2020-02-21
Olp 665

kingpinXD:


action/init.go
```diff
@@ -42,6 +42,7 @@ const (
 	ETH_REDEEM               Type = 0x96
 	ERC20_LOCK               Type = 0x97
 	ERC20_REDEEM             Type = 0x98
+	ETH_REPORT_FAILED        Type = 0x99
```
kingpinXD: Decided not to use a seperate TX type , and using the refund flag in the existing ReportFinality mint struct.  Will rename the refund to a different identifier

data/ethereum/init.go
```diff
@@ -9,12 +9,14 @@ const (
...
 	Finalized
 	Released
+	Failed
 
 	BROADCASTING  string = "broadcasting"
 	FINALIZING    string = "finalizing"
 	FINALIZE      string = "finalize"
 	MINTING       string = "minting"
 	CLEANUP       string = "cleanup"
+	CLEANUPFAILED string = "cleanupfailed"
```
kingpinXD: This cleanup function , only deletes jobs if they exist . If they dont it does not cause a run-time failure from storage

data/ethereum/init.go
```diff
@@ -50,4 +52,24 @@ func GetProcessTypeString(t ProcessType) string {
 	return "UNKNOWN TYPE"
 }
 
+func GetTrackerStateString(t TrackerState) string {
+	switch t {
+	case 0:
+		return "NEWTRACKER"
```
alexjlan: make ToString func for TrackerState

event/report_failed.go
```diff
@@ -0,0 +1,51 @@
...
+	"errors"
+
+	"github.com/Oneledger/protocol/action"
+	"github.com/Oneledger/protocol/action/eth"
+	"github.com/Oneledger/protocol/chains/ethereum"
+)
+
+func BroadcastReportFailedETHTx(ctx interface{},trackerName ethereum.TrackerName ,jobID string) (error){
+	ethCtx, _ := ctx.(*JobsContext)
+	trackerStore := ethCtx.EthereumTrackers
```
alexjlan: make this function handle positive cases and don't pass interface{}

PR #492, opened 2020-04-29
Develop

alexjlan:


scripts/ons/get_all_domains.py
```diff
@@ -0,0 +1,10 @@
+import time
+import sys
```
ws4charlie, line 2: Are we going to stay with Python 2 for SDK

kingpinXD:
we have it here 
https://oneledger.atlassian.net/wiki/spaces/EN/pages/532054112/Mainneet+Launch+Files
Will probbaly keep it in the VM where we run the command to generate all the nodes

PR #500, opened 2020-05-11
Master fix

kingpinXD:


scripts/ethereum/main.go
```diff
@@ -48,7 +46,7 @@ var (
 	TestTokenABI     = contract.ERC20BasicABI
 	LockRedeemERCABI = contract.LockRedeemERCABI
 	// LockRedeemERC20ABI = contract.ContextABI
-	LockRedeemContractAddr      = "0xd2629c4282E2946a6d762120164887e8f48C6AeB"
+	LockRedeemContractAddr      = "0xDc9F35C56344A769D4908044Ab35AE00Faa01f1A"
```
pav-kap, line 49: why did u change the smart contract address?

scripts/resetDev
```diff
@@ -7,9 +7,9 @@ rm -rf $OLDATA/devnet/*
 olfullnode init devnet --dir $OLDATA/devnet --nonvalidators 1 \
 --empty_blocks \
 --loglevel 4 \
---eth_rpc HTTP://127.0.0.1:7545 \
-#--total_funds 1000000000 \
-#--deploy_smart_contracts \
+--eth_rpc https://rinkeby.infura.io/v3 \
```
pav-kap, line 10: Are these changes  needed?

cmd/olfullnode/devnet.go
```diff
@@ -600,7 +600,7 @@ func deployethcdcontract(conn string, nodeList []node) (*ethchain.ChainDriverOpt
 	lock_period := big.NewInt(25)
 
 	tokenSupplyTestToken := new(big.Int)
-	validatorInitialFund := big.NewInt(300000000000000000) //300000000000000000
+	validatorInitialFund := big.NewInt(0) //300000000000000000
```
pav-kap, line 603: Is this needed?

event/eth_sign_redeem.go
```diff
@@ -121,21 +121,30 @@ func (j *JobETHSignRedeem) DoMyJob(ctx interface{}) {
...
+		BroadcastReportFinalityETHTx(ctx.(*JobsContext), j.TrackerName, j.JobID, false)
+	}
+	//Signature confirmed
 	if success {
 		ethCtx.Logger.Debug("validator Sign Confirmed | validator Address (SIGNER):", ethCtx.GetValidatorETHAddress().Hex(), "| User Eth Address :", msg.From().Hex())
 		j.Status = jobs.Completed
 		return
 	}
+
 	if j.RetryCount >= 0 && !success {
```
pav-kap, line 138: No point in checking !success.. as it will always be true at this place. You can try to move this condition above , if that makes sense

pav-kap: Or probably check for err != nil

scripts/ethereum/main.go
```diff
@@ -365,18 +363,18 @@ func redeem() []byte {
 		return nil
 	}
 
-	//acc := accReply.Accounts[0]
+	acc := accReply.Accounts[0]
```
pav-kap, line 366: Are these changes related to this pr?

chains/ethereum/online_chain_driver.go
```diff
@@ -136,7 +136,7 @@ func (acc *ETHChainDriver) SignRedeem(fromaddr common.Address, redeemAmount *big
 	if err != nil {
 		return nil, err
 	}
-	gasLimit := uint64(6721974)
+	gasLimit := uint64(700000)
```
pav-kap, line 139: Can we add this as a constant somewhere.. so that we dont need to change it at multiple places?

event/eth_sign_redeem.go
```diff
@@ -87,20 +87,20 @@ func (j *JobETHSignRedeem) DoMyJob(ctx interface{}) {
 	tx, err := cd.DecodeTransaction(rawTx)
 	if err != nil {
 		ethCtx.Logger.Error("Error Decoding Bytes from RaxTX :", j.GetJobID(), err)
-		return
+		//return
```
pav-kap, line 90: This has to be implemented thoroughly.. I m not sure if its good idea to remove the return here. Which means, even though the decode fails, it will go to line 93 to get the transaction message. Is that a good idea?

If the oETH is already burnt in these cases, then we should refund in all these cases, instead of just returning here...

You also need to consider the situation of lock here...Not just redeem

event/eth_sign_redeem.go
```diff
@@ -87,20 +87,20 @@ func (j *JobETHSignRedeem) DoMyJob(ctx interface{}) {
...
 	}
 
 	msg, err := cd.GetTransactionMessage(tx)
 	if err != nil {
 		ethCtx.Logger.Error("Error in decoding transaction as message : ", j.GetJobID(), err)
-		return
+		//return
```
pav-kap, line 96: same as above

event/eth_sign_redeem.go
```diff
@@ -87,20 +87,20 @@ func (j *JobETHSignRedeem) DoMyJob(ctx interface{}) {
...
 	}
 
 	addr := ethCtx.GetValidatorETHAddress()
 	txReceipt, err := cd.VerifyReceipt(tx.Hash())
 	if err != nil {
 		ethCtx.Logger.Debug("Trying to confirm RedeemTX sent by User Receipt :", err)
-		return
+		//return
```
pav-kap, line 103: same as above

client/request_types.go
```diff
@@ -10,6 +10,7 @@
 package client
```
pav-kap, line 10: Why does this commit have all these changes?
@kingpinXD are you sure your pr is right?

chains/ethereum/online_chain_driver.go
ws4charlie: Constant?

service/query/service.go
```diff
@@ -31,14 +32,15 @@ func Name() string {
 	return "query"
 }
 
-func NewService(ctx client.ExtServiceContext, balances *balance.Store, currencies *balance.CurrencySet, validators *identity.ValidatorStore,
+func NewService(ctx client.ExtServiceContext, balances *balance.Store, currencies *balance.CurrencySet, validators *identity.ValidatorStore, witnesses *identity.WitnessStore,
```
ws4charlie, line 35: Is this new listWitness service related to the fix of ETH Redeem?

scripts/resetDev
ws4charlie: same question as Pavan.

scripts/ethereum/main.go
ws4charlie: Are these changes only related to integration/unit tests?

scripts/resetDev
ws4charlie: Only for devnet/local test, right?

app/context.go
```diff
@@ -209,6 +209,7 @@ func (ctx *context) Services() (service.Map, error) {
 		Cfg:          ctx.cfg,
 		NodeContext:  ctx.node,
 		ValidatorSet: identity.NewValidatorStore("v", storage.NewState(ctx.chainstate)),
+		WitnessSet:   identity.NewWitnessStore("w", storage.NewState(ctx.chainstate)),
```
ws4charlie, line 212: Is this related to the fix or not?

cmd/olfullnode/devnet.go
ws4charlie: Should this PR be made from **MASTER** to the deployed taged version?

PR #502, opened 2020-05-11
Redeem hotfix

kingpinXD:
Tests 
Tested with old binary , Issue with 
- Insufficient funds for validator replicated
- Issue with redeem expiring before validator signature is mine replicated

Tested with fined binary 
- Tracker failer in both cases
- Oeth Refunded in both cases  
- Some validators loose connection to infura and panic .

more than 33% validators loose connection ,and panic
before we bring the nodes back up ,redeem expires ,and  validators cast there No votes ,
After Casting to no vote some of those validators also restart ,becuase of some reason
so there no votes are still in sessioncache and not chainstate ,when they come back up they wont have jobs fo signing ,and the tracker woud be missing that many number of votes

### DIFF
--- a/app/application.go
+++ b/app/application.go
@@ -371,19 +371,30 @@ func (app *App) Prepare() error {
 
 // Start initializes the state
 func (app *App) Start() error {
-	app.logger.Info("Starting node...")
 
 	err := app.Prepare()
 	if err != nil {
 		return err
 	}
+	// Adding internal Router
+	internalRouter := action.NewRouter("internal")
+	err = eth.EnableInternalETH(internalRouter)
+	if err != nil {
+		app.logger.Error("failed to register eth internal transaction")
+		return err
+	}
+	app.Context.internalService = event.NewService(app.Context.node,
+		log.NewLoggerWithPrefix(app.Context.logWriter, "internal_service"), internalRouter, app.node)
 
+	// Starting App
 	err = app.node.Start()
 	if err != nil {
 		app.logger.Error("Failed to start consensus.Node")
 		return errors.Wrap(err, "failed to start new consensus.Node")
 	}
-
+	//Start Jobbus
+	_ = app.Context.jobBus.Start(app.Context.JobContext())
+	// Starting RPC
 	startRPC, err := app.rpcStarter()
 	if err != nil {
 		return errors.Wrap(err, "failed to prepare rpc service")
@@ -395,23 +406,12 @@ func (app *App) Start() error {
 		return err
 	}
 
-	internalRouter := action.NewRouter("internal")
 	//"btc" service temporarily disabled
 	//err = btc.EnableBTCInternalTx(internalRouter)
 	//if err != nil {
 	//	app.logger.Error("Failed to register btc internal transactions")
 	//	return err
 	//}
-	err = eth.EnableInternalETH(internalRouter)
-	if err != nil {
-		app.logger.Error("failed to register eth internal transaction")
-		return err
-	}
-
-	app.Context.internalService = event.NewService(app.Context.node,
-		log.NewLoggerWithPrefix(app.Context.logWriter, "internal_service"), internalRouter, app.node)
-
-	_ = app.Context.jobBus.Start(app.Context.JobContext())
 
 	return nil
 }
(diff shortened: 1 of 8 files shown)


event/eth_sign_redeem.go
```diff
@@ -97,60 +97,74 @@ func (j *JobETHSignRedeem) DoMyJob(ctx interface{}) {
...
-		return
-	}
+	//Signature confirmed
 	if success {
 		ethCtx.Logger.Debug("validator Sign Confirmed | validator Address (SIGNER):", ethCtx.GetValidatorETHAddress().Hex(), "| User Eth Address :", msg.From().Hex())
 		j.Status = jobs.Completed
 		return
 	}
+	//Log print debugger
 	if j.RetryCount >= 0 && !success {
```
pav-kap, line 134: !success will always be true here.

chains/ethereum/types.go
```diff
@@ -13,10 +13,11 @@ import (
 type RedeemStatus int8
 
 const (
-	NewRedeem RedeemStatus = -1
-	Ongoing   RedeemStatus = 0
-	Success   RedeemStatus = 1
-	Expired   RedeemStatus = 2
+	NewRedeem       RedeemStatus = -1
```
pav-kap, line 16: Good job of maintaining these constants

event/eth_verify_redeem.go
```diff
@@ -71,19 +71,20 @@ func (job *JobETHVerifyRedeem) DoMyJob(ctx interface{}) {
...
-	status, err := cd.VerifyRedeem(addr, msg.From())
-	if err != nil {
-		ethCtx.Logger.Error("Error in verifying redeem :", job.GetJobID(), err)
+	status := cd.VerifyRedeem(addr, msg.From())
+	if status == ethereum.ErrorConnecting {
+		ethCtx.Logger.Error("Unable to connect to ethereum smartcontract")
+		return
 	}
-	if err == ethereum.ErrRedeemExpired {
+	if status == ethereum.Expired {
```
pav-kap, line 79: Is this the only case where user gets his oETH refunded? Did u take care of other cases?

chains/ethereum/online_chain_driver.go
```diff
@@ -136,7 +136,7 @@ func (acc *ETHChainDriver) SignRedeem(fromaddr common.Address, redeemAmount *big
 	if err != nil {
 		return nil, err
 	}
-	gasLimit := uint64(6721974)
+	gasLimit := uint64(700000)
```
ws4charlie: do you want to make it as Constant?

event/eth_sign_redeem.go
ws4charlie: @pav-kap I think what you meat was that it's pointless to check the flag success when error is not nil, right?

ws4charlie: @pav-kap if that's the case, we can make the above err check as 
if err != nil or !success {
    ethCtx.Logger.Error("some error msg")
}

pav-kap: I mean if success is true, it should have already went inside if success {} block, and returned.. Only since its not success, it came down.. So !success is always true here

chains/ethereum/online_chain_driver.go
```diff
@@ -162,7 +162,7 @@ func (acc *ETHChainDriver) PrepareUnsignedETHLock(addr common.Address, lockAmoun
...
```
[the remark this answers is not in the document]
kingpinXD: Added a constant , Might thread it through genesis later

event/eth_sign_redeem.go
kingpinXD: Removed the extra flag

chains/ethereum/types.go
```diff
@@ -29,6 +30,8 @@ func (r RedeemStatus) String() string {
 		return "Success"
 	case 2:
 		return "Expired"
+	case 3:
```
ws4charlie, line 33: is this one -2 or 3?

kingpinXD: Can be anything , just added -2 to signify that its different , the rest of status are returned directly from smart contract

app/application.go
```diff
@@ -371,19 +371,30 @@ func (app *App) Prepare() error {
 
 // Start initializes the state
 func (app *App) Start() error {
-	app.logger.Info("Starting node...")
 
 	err := app.Prepare()
 	if err != nil {
 		return err
 	}
+	// Adding internal Router
```
alexjlan, line 379: should move this part into Prepare

kingpinXD: Moved

PR #503, opened 2020-05-13
split the witness and validator totally

alexjlan:


scripts/ethereum/main.go
```diff
@@ -48,7 +46,7 @@ var (
 	TestTokenABI     = contract.ERC20BasicABI
 	LockRedeemERCABI = contract.LockRedeemERCABI
 	// LockRedeemERC20ABI = contract.ContextABI
-	LockRedeemContractAddr      = "0xd2629c4282E2946a6d762120164887e8f48C6AeB"
```
pav-kap, line 51: Is this change needed?

scripts/ethereum/main.go
```diff
@@ -235,21 +233,21 @@ func lock() []byte {
...
-	err = acc.UnmarshalText([]byte("0x416e9cc0abc4ea98b4066823a62bfa6515180582"))
-	if err != nil {
-		return nil
-	}
-	wallet, err := accounts.NewWalletKeyStore("/home/tanmay/Codebase/Test/WalletStore/keystore")
-	if err != nil {
-		return nil
-	}
-	wallet.Open(acc, "123")
+	acc := accReply.Accounts[0]
```
pav-kap, line 236: Looks like unrelated changes

cmd/olfullnode/save_state.go
```diff
@@ -336,6 +340,31 @@ func DumpFeesToFile(st *fees.Store, writer io.Writer, fn func(writer io.Writer,
 	return
 }
 
+func DumpWitnessToFile(ws *identity.WitnessStore, writer io.Writer, fn func(writer io.Writer, obj interface{}) bool) {
```
pav-kap, line 343: I would make a ethereum utility file, and add all these ethereum specific changes in that file. In this place, I would iterate all the chain specific utility files, and call the same DumpWitnessToFile method. That way when we add a new chain of witnesses, we dont need to change at multiple places

scripts/ethereum/main.go
alexjlan: this is the test scripts for ethereum lock and redeem, I merged it from Tanmay's changes so I can test.

scripts/ethereum/main.go
alexjlan: it's test scripts

cmd/olfullnode/save_state.go
alexjlan: I'd rather do that when we have bitcoin witness, since we are not sure the witness object will change or not for the btc.

alexjlan: and this is the only place that dump for witness is call and used, if we create another file, we actually need to change in multiple place, and this function is update only when new chain is added for witness.

cmd/olfullnode/devnet.go
```diff
@@ -468,16 +482,30 @@ func initialState(args *testnetConfig, nodeList []node, option ethchain.ChainDri
...
+				StakeAddress:     stakeAddr,
+				Pubkey:           pubkey,
+				ECDSAPubKey:      h.PubKey(),
+				Name:             node.validator.Name,
+				Amount:           *vt.NewCoinFromInt(node.validator.Power).Amount,
+			}
+			staking = append(staking, st)
+		}
+
+		if node.isWitness {
```
pav-kap, line 498: Are we sure we don't want to give a new token (something like WT) for witnesses. I understand that we don't need that right now. But probably it's a good idea to add a new dummy token WT or ( EWT, Ethereum Witness Token) for Eth witnesses, so that we can think of making It public down the line, somewhere in future

cmd/olfullnode/genesis.go
```diff
@@ -76,6 +78,8 @@ func init() {
 	//genesisCmd.Flags().BoolVar(&genesisCmdArgs.allowSwap, "enable_swaps", false, "Allow swaps")
 	genesisCmd.Flags().IntVar(&genesisCmdArgs.numValidators, "validators", 4, "Number of validators to initialize mainnetnet with")
 	genesisCmd.Flags().IntVar(&genesisCmdArgs.numNonValidators, "nonvalidators", 1, "Number of fullnodes to initialize mainnetnet with")
+	genesisCmd.Flags().IntVar(&genesisCmdArgs.numWitness, "witness", 0, "number of witness to initialize mainnet with")
```
pav-kap, line 81: I want to call it ethereum withness, instead of witness here... Just calling it witness, we are adding more work when other chains are integrated..

cmd/olfullnode/devnet.go
```diff
@@ -85,6 +86,8 @@ func init() {
 	initCmd.AddCommand(testnetCmd)
 	testnetCmd.Flags().IntVar(&testnetArgs.numValidators, "validators", 4, "Number of validators to initialize devnet with")
 	testnetCmd.Flags().IntVar(&testnetArgs.numNonValidators, "nonvalidators", 0, "Number of non-validators to initialize the devnet with")
+	testnetCmd.Flags().IntVar(&testnetArgs.numWitness, "witness", 0, "number of witness to initialize mainnet with")
```
pav-kap, line 89: Im sorry, but Im totally confused by the usage of the word witness... I feel this is not easily extendable to other chains like BTC or even any ERC20 token. With these changes, I sort of expected the addition of new chains would be a breeze and doesn't need a lot of work again changing a lot of these files.

app/context.go
```diff
@@ -248,6 +248,7 @@ type StorageCtx struct {
 	Balances   *balance.Store
 	Domains    *ons.DomainStore
 	Validators *identity.ValidatorStore // Set of validators currently active
+	Witnesses  *identity.WitnessStore
```
kingpinXD, line 251: We need to add witnesses to the Job context as well  ?
We can rename the JobContext as Ethereum and Bitcoin job context ,and separate out the events package ?

alexjlan:
so no one is merging this branch?

PR #507, opened 2020-05-20
OLP-947 Query All Second Level Domain(parent domain) by Owner Address

kerhbal:
created a new query domain service that only returns parent domain. 
and some python integration test changes

### DIFF
--- a/service/query/ons.go
+++ b/service/query/ons.go
@@ -55,6 +55,35 @@ func (sv *Service) ONS_GetDomainByOwner(req client.ONSGetDomainsRequest, reply *
 	return nil
 }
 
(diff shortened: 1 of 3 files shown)


service/query/ons.go
```diff
@@ -55,6 +55,35 @@ func (sv *Service) ONS_GetDomainByOwner(req client.ONSGetDomainsRequest, reply *
...
+				return false
+			}
+			ds = append(ds, *domain)
+		}
+		return false
+	})
+
+	*reply = client.ONSGetDomainsReply{
+		Domains: ds,
+		Height:  sv.balances.State.Version(),
```
KevinOneLedger, line 81: I think you meant to use:
sv.domains.State.Version()

kerhbal: yeah, I'll also change other domain query, since they share same problems.

kerhbal: actually it's sv.ons.State.Version()

PR #520, opened 2020-05-27
Self staking mechanizm

DeV1doR:


Makefile
```diff
@@ -56,30 +56,13 @@ utest:
 		github.com/Oneledger/protocol/rpc \
 		github.com/Oneledger/protocol/identity \
 		github.com/Oneledger/protocol/app \
+		github.com/Oneledger/protocol/action/staking \
```
ws4charlie, line 59: Writing unit tests in here takes lots of time on setting up the environment. We are now testing Txs by using integration tests located in protocol/scripts/ folder

action/base.go
```diff
@@ -78,6 +79,9 @@ func (t *SignedTx) SignedBytes() []byte {
 }
 
 func ValidateBasic(data []byte, signerAddr []Address, signatures []Signature) error {
+	if len(signatures) < len(signerAddr) {
```
ws4charlie, line 82: Should exactly match, right? Shall we replace < with !=  ?

action/init.go
```diff
@@ -13,9 +13,9 @@ const (
 	SEND Type = 0x01
 
 	//staking related transaction
-	APPLYVALIDATOR Type = 0x11
```
ws4charlie, line 16: Since we keep committing new changes to develop branch, It's good for you to keep your branch sync with develop because conflicts could be painful in final merge.

action/base.go
DeV1doR: Not exactly as u can have a required signers (as 2 for example), but u could put 3 signatures for check

action/init.go
DeV1doR: No pain of rebasing 5-10 minutes took previous time. APPLYVALIDATOR works 50 vs 50, so in some execution the node could stop after this action.

Makefile
DeV1doR: How this is related to integration tests?

action/staking/stake.go
```diff
@@ -0,0 +1,163 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	err = action.ValidateFee(ctx.FeePool.GetOpt(), tx.Fee)
+	if err != nil {
+		return false, err
+	}
+
+	if len(st.StakeAddress) == 0 {
```
ws4charlie, line 76: You can use a 'Err()' method defined in address struct which perfectly helps checking the address format.

action/staking/stake.go
```diff
@@ -0,0 +1,163 @@
...
+	coin := st.Stake.ToCoinWithBase(ctx.Currencies)
+
+	err = ctx.Balances.MinusFromAddress(st.StakeAddress, coin)
+	if err != nil {
+		return false, action.Response{Log: errors.Wrap(err, st.StakeAddress.String()).Error()}
+	}
+
+	err = ctx.Delegators.Stake(st.ValidatorAddress, st.StakeAddress, st.Stake.Value)
+	if err != nil {
+		return false, action.Response{Log: errors.Wrap(err, st.StakeAddress.String()).Error()}
```
ws4charlie, line 154: I think we need to add the coin back to stake address, the money will get lost otherwise. Make sure we raise a panic if we failed to rollback the money to stake address, since Alex said node can safely handle panic in Tx.

action/staking/stake.go
```diff
@@ -0,0 +1,163 @@
...
+	}
+
+	err = ctx.Validators.HandleStake(stake)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
```
ws4charlie, line 159: We reorganized and unified errors and error codes since we introduced governance. In folder status_codes/codes.go, we defined the error codes needed in our transactions. In folder identity/, you can create a new file 'errors.go' and put all error objects there, like data/balance/errors.go

action/staking/unstake.go
```diff
@@ -0,0 +1,138 @@
...
+	if err != nil {
+		return false, err
+	}
+
+	err = action.ValidateFee(ctx.FeePool.GetOpt(), tx.Fee)
+	if err != nil {
+		return false, err
+	}
+
+	if len(ust.StakeAddress) == 0 {
```
ws4charlie, line 73: The same Err() method can help checking here.

action/staking/unstake.go
```diff
@@ -0,0 +1,138 @@
...
+		return false, action.Response{Log: errors.Wrap(err, ust.StakeAddress.String()).Error()}
+	}
+
+	err = ctx.Delegators.Unstake(ust.ValidatorAddress, ust.StakeAddress, ust.Stake.Value, height+options.MaturityTime)
+	if err != nil {
+		return false, action.Response{Log: errors.Wrap(err, ust.StakeAddress.String()).Error()}
+	}
+
+	err = ctx.Validators.HandleUnstake(unstake)
+	if err != nil {
```
ws4charlie, line 133: What happens if Delegators.Unstake succeed and Validators.HandleUnstake faile? Can we roll-back or raise panic?

action/staking/withdraw.go
```diff
@@ -42,14 +43,10 @@ func (s Withdraw) Tags() kv.Pairs {
...
-		Value: s.From.Bytes(),
-	}
-	tag3 := kv.Pair{
-		Key:   []byte("tx.to"),
-		Value: s.To.Bytes(),
+		Value: s.StakeAddress.Bytes(),
 	}
 
-	tags = append(tags, tag, tag2, tag3)
+	tags = append(tags, tag, tag2)
```
ws4charlie, line 49: Can we add full information to tags(validator address and amount), explorer might need to catch and parse information from this event message and save to Postgresql database.

action/staking/withdraw.go
```diff
@@ -59,29 +56,38 @@ type withdrawTx struct {
...
-	feeOpt := ctx.FeePool.GetOpt()
-	err = action.ValidateFee(feeOpt, tx.Fee)
+	err = action.ValidateFee(ctx.FeePool.GetOpt(), tx.Fee)
 	if err != nil {
 		return false, err
 	}
 
-	if tx.Fee.Price.Currency != feeOpt.FeeCurrency.Name {
-		return false, action.ErrInvalidFeeCurrency
+	if len(apply.StakeAddress) == 0 {
```
ws4charlie, line 76: Same Error() method to check.

action/staking/unstake.go
DeV1doR: If we receive an error from ProcessCheck and/or ProcessDeliver, we will discard this transaction:
https://github.com/Oneledger/protocol/blob/master/app/controller.go#L238
https://github.com/Oneledger/protocol/blob/master/app/controller.go#L183

action/staking/stake.go
DeV1doR: If we receive an error from ProcessCheck and/or ProcessDeliver, we will discard this transaction:
https://github.com/Oneledger/protocol/blob/master/app/controller.go#L238
https://github.com/Oneledger/protocol/blob/master/app/controller.go#L183

Why we need a panic if it will be a false and we will revert it?

action/staking/stake.go
DeV1doR: Code is ok, but as a developer I want to understand directly where an error (even when I use API). In case of HandleUnstake it manipulates with 3 different keys, and different errors could take a place. As a normal user, he must not receive an error in this place (only when he will try something to hack with storage)

action/staking/withdraw.go
DeV1doR: Added

cmd/olclient/stake.go
```diff
@@ -0,0 +1,140 @@
...
+	}
+
+	signedTx.Signatures = append([]action.Signature{{Signer: pub, Signed: signature}}, signedTx.Signatures...)
+
+	packet, err := serialize.GetSerializer(serialize.NETWORK).Serialize(signedTx)
+	if packet == nil || err != nil {
+		return errors.New("error serializing packet: " + err.Error())
+	}
+
+	result, err := ctx.clCtx.BroadcastTxCommit(packet)
```
kingpinXD, line 132: Can we use BroadcastTXSync instead ? Commit might lead to a timeout waiting for block to be included .

cmd/olclient/unstake.go
```diff
@@ -0,0 +1,136 @@
...
+	}
+
+	signedTx.Signatures = append([]action.Signature{{Signer: pub, Signed: signature}}, signedTx.Signatures...)
+
+	packet, err := serialize.GetSerializer(serialize.NETWORK).Serialize(signedTx)
+	if packet == nil || err != nil {
+		return errors.New("error serializing packet: " + err.Error())
+	}
+
+	result, err := ctx.clCtx.BroadcastTxCommit(packet)
```
kingpinXD, line 128: same as stake

service/query/service.go
```diff
@@ -80,9 +86,44 @@ func (svc *Service) ListValidators(_ client.ListValidatorsRequest, reply *client
...
+	if err != nil {
+		svc.logger.Error("error to fetch staking options")
+		return codes.ErrListValidators
+	}
+
+	vMap := make(map[string]bool)
+	cnt := int64(0)
+
+	i := 0
+	queue := &identity.ValidatorQueue{PriorityQueue: make(utils.PriorityQueue, 0, 100)}
```
KevinOneLedger, line 99: Why do you need to use the queue to store the validators? Is there a reason you can't set them directly to vMap?

data/governance/store.go
```diff
@@ -152,6 +154,37 @@ func (st *Store) SetEpoch(epoch int64) error {
...
+func (st *Store) SetStakingOptions(opt delegation.Options) error {
+
+	bytes, err := serialize.GetSerializer(serialize.PERSISTENT).Serialize(opt)
+	if err != nil {
+		return errors.Wrap(err, "failed to serialize eth chaindriver option")
+	}
+
+	err = st.Set([]byte(ADMIN_STAKING_OPTION), bytes)
+	if err != nil {
+		return errors.Wrap(err, "failed to set the eth chaindriver option")
```
kingpinXD, line 182: I think there might have been a typo here

action/base.go
KevinOneLedger: The signers and signatures are supposed to be in order, that is how it gets verified.

data/delegation/store.go
```diff
@@ -0,0 +1,391 @@
...
+
+	prefixKey := append(st.prefix, key...)
+	err = st.state.Set(storage.StoreKey(prefixKey), dat)
+	return err
+}
+
+// Validator - Delegator data
+
+func (st *DelegationStore) getVDKey(validatorAddress keys.Address, delegatorAddress keys.Address) []byte {
+	key := []byte(fmt.Sprintf("_e_%s_%s", validatorAddress, delegatorAddress))
```
ws4charlie, line 70: Can we use String() method to convert address to string? Because printing addresses as %s is not human readable, and this conversion helps trouble shooting.

data/delegation/store.go
```diff
@@ -0,0 +1,391 @@
...
+	Data   []*MatureData
+}
+
+type MatureData struct {
+	Address keys.Address
+	Amount  balance.Amount
+	Height  int64
+}
+
+func NewDelegationStore(prefix string, state *storage.State) *DelegationStore {
```
ws4charlie, line 31: Add unit tests at store level?

data/delegation/store.go
```diff
@@ -0,0 +1,391 @@
...
+	return &DelegationStore{
+		state:  state,
+		prefix: storage.Prefix(prefix),
+		szlr:   serialize.GetSerializer(serialize.PERSISTENT),
```
ws4charlie, line 35: This szlr is indented to replace the invoking of "serialize.GetSerializer(serialize.PERSISTENT)" but actually never used.

data/delegation/store.go
```diff
@@ -0,0 +1,391 @@
...
+func (st *DelegationStore) SetValidatorDelegationAmount(validatorAddress keys.Address, delegatorAddress keys.Address, amt balance.Amount) (err error) {
+	key := st.getVDKey(validatorAddress, delegatorAddress)
+	err = st.Set(key, &amt)
+	return
+}
+
+// Validator data
+
+func (st *DelegationStore) getVKey(validatorAddress keys.Address) []byte {
+	key := []byte(fmt.Sprintf("_t_%s", validatorAddress))
```
ws4charlie, line 89: Same as above

data/delegation/store.go
```diff
@@ -0,0 +1,391 @@
...
+
+type DelegationStore struct {
+	state  *storage.State
+	szlr   serialize.Serializer
+	prefix []byte
+	mux    sync.Mutex
+}
+
+type MatureBlock struct {
+	Height int64
```
ws4charlie, line 21: We keep this Height field but seems it's used no where.

action/staking/stake.go
```diff
@@ -0,0 +1,171 @@
...
+	tag3 := kv.Pair{
+		Key:   []byte("tx.delegator"),
+		Value: st.StakeAddress.Bytes(),
+	}
+	tag4 := kv.Pair{
+		Key:   []byte("tx.amount"),
+		Value: st.Stake.Value.BigInt().Bytes(),
+	}
+
+	tags = append(tags, tag, tag2, tag3, tag4)
```
ws4charlie, line 60: Put all fields to tags?

action/staking/stake.go
```diff
@@ -0,0 +1,171 @@
...
+	err = action.ValidateFee(ctx.FeePool.GetOpt(), tx.Fee)
+	if err != nil {
+		return false, err
+	}
+
+	if err := st.StakeAddress.Err(); err != nil {
+		return false, err
+	}
+
+	if st.ValidatorAddress == nil {
```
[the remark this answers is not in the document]
ws4charlie: Yes. It's good to unify.

action/staking/withdraw.go
```diff
@@ -59,29 +64,38 @@ type withdrawTx struct {
...
 		return false, err
 	}
 
-	if tx.Fee.Price.Currency != feeOpt.FeeCurrency.Name {
-		return false, action.ErrInvalidFeeCurrency
+	if err := apply.StakeAddress.Err(); err != nil {
+		return false, err
+	}
+
+	if apply.ValidatorAddress == nil {
```
ws4charlie, line 88: Better to use Err() to check as above.

action/staking/stake.go
DeV1doR: NodeName, ValidatorECDSAPubKey and ValidatorPubKey - redundant and could be taken from leveldb.

identity/validator_set.go
```diff
@@ -63,6 +64,10 @@ func (vs *ValidatorStore) Exists(addr keys.Address) bool {
 	return vs.store.Exists(key)
 }
 
+func (vs *ValidatorStore) Set(validator Validator) error {
```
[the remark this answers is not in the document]
DeV1doR: Why it must be private?

data/delegation/store.go
DeV1doR: Covered during stake, unstake, withdraw tests

data/delegation/store.go
DeV1doR: https://github.com/Oneledger/protocol/blob/4cb6108ed75db4426c4407a40f931ad19b7ebeca/data/delegation/store.go#L156
https://github.com/Oneledger/protocol/blob/4cb6108ed75db4426c4407a40f931ad19b7ebeca/identity/validator_set.go#L370

cmd/olclient/stake.go
DeV1doR: https://godoc.org/github.com/tendermint/tendermint/rpc/core#BroadcastTxSync
Does not wait for DeliverTx result. But we want to be make sure that it was broadcasted.

cmd/olclient/unstake.go
DeV1doR: answer at cmd/olclient/stake.go

action/base.go
DeV1doR: I can remove, but if signers will be mismatched u will get a panic at this line:
pkey := signatures[i].Signer

if you don't like this solution, then propose another one. But anyway, if I have a 1 signer and we got a multiple signatures, how It be resolved?

cmd/olclient/stake.go
kingpinXD: Agreed , but commit might not always be a success when using a VMs in a bigger network , we can use subscribe to the TX result via the web-socket if needed

action/staking/stake.go
```diff
@@ -0,0 +1,163 @@
...
+		return false, errors.Wrap(action.ErrInvalidAmount, coin.String())
+	}
+
+	if coin.Currency.Name != "OLT" {
+		return false, action.ErrInvalidCurrency
+	}
+
+	_, ok := ctx.Currencies.GetCurrencyByName("OLT")
+	if !ok {
+		return false, action.ErrInvalidCurrency
```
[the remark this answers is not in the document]
DeV1doR: Why if transaction will be failed as we return false?

data/delegation/store.go
DeV1doR: `fmt.Println(fmt.Sprintf("_e_%s_%s", validatorAddress, delegatorAddress))`
`_e_0lt908566b0c536e011ae4f4c25f67bf3c55ef38a26_0lt9d35e060a5fddcc200551d5b4a70efabb31483bb`

DeV1doR: seems like it human readable as Address has implementation of String()

PR #522, opened 2020-05-29
Olp 904

kerhbal:
https://oneledger.atlassian.net/browse/OLP-904

### DIFF
--- a/action/governance/withdrawFunds.go
+++ b/action/governance/withdrawFunds.go
@@ -134,19 +134,9 @@ func runWithdraw(ctx *action.Context, signedTx action.RawTx) (bool, action.Respo
 		return false, result
 	}
 
-	// 4. Check if the contributor has sufficient funds to withdraw for that proposal
+	// 4. withdraw
+	// deduct from proposal fund and check if the contributor has sufficient funds to withdraw for that proposal
 	withdrawAmount := balance.NewAmountFromBigInt(withdrawProposal.WithdrawValue.Value.BigInt())
-	//if proposalFund.BigInt().Cmp(withdrawAmount.BigInt()) < 0 {
-	//	ctx.Logger.Error("Insufficient funds to withdraw for this contributor :", withdrawProposal.Contributor)
-	//	result := action.Response{
-	//		Events: action.GetEvent(withdrawProposal.Tags(), "no_available__fund_to_withdraw_for_this_contributor"),
-	//		Log: action.ErrNotEnoughFund.Msg,
-	//	}
-	//	return false, result
-	//}
-
-	// 5. withdraw
-	// deduct from proposal fund
 	err = ctx.ProposalMasterStore.ProposalFund.DeductFunds(proposal.ProposalID, withdrawProposal.Contributor, withdrawAmount)
 	if err != nil {
 		ctx.Logger.Error("Failed to deduct funds from proposal:", withdrawProposal.ProposalID)
--- a/action/governance/withdrawFunds.go
+++ b/action/governance/withdrawFunds.go
@@ -106,7 +106,6 @@ func runWithdraw(ctx *action.Context, signedTx action.RawTx) (bool, action.Respo
 		return false, result
 	}
 	// 2. change outcome, status, state
-	// todo not sure this part should be done here or in blockender
 	proposal.Outcome = governance.ProposalOutcomeInsufficientFunds
 	proposal.Status = governance.ProposalStatusCompleted
 	err = ctx.ProposalMasterStore.Proposal.WithPrefixType(governance.ProposalStateFailed).Set(proposal)
@@ -142,6 +141,7 @@ func runWithdraw(ctx *action.Context, signedTx action.RawTx) (bool, action.Respo
 		ctx.Logger.Error("Insufficient funds to withdraw for this contributor :", withdrawProposal.Contributor)
 		result := action.Response{
 			Events: action.GetEvent(withdrawProposal.Tags(), "no_available__fund_to_withdraw_for_this_contributor"),
+			Log: action.ErrNotEnoughFund.Msg,
 		}
 		return false, result
 	}
@@ -152,7 +152,7 @@ func runWithdraw(ctx *action.Context, signedTx action.RawTx) (bool, action.Respo
 	if err != nil {
 		ctx.Logger.Error("Failed to deduct funds from proposal:", withdrawProposal.ProposalID)
 		result := action.Response{
-			Events: action.GetEvent(withdrawProposal.Tags(), "withdraw_proposal_DeductFund_failed"),
+			Events: action.GetEvent(withdrawProposal.Tags(), "withdraw_proposal_deduct_fund_failed"),
 		}
 		return false, result
 	}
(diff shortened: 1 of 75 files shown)


client/gov_request_types.go
```diff
@@ -33,6 +34,7 @@ type GetProposalByIDRequest struct {
 
 type GetProposalByIDReply struct {
```
ws4charlie, line 35: I modified the struct's name and fields for 'olclient' command. Let's resolve our conflicts before merge.

data/governance/proposal_fund.go
```diff
@@ -31,3 +31,21 @@ func GetCurrentFunds(id ProposalID, store *ProposalFundStore) *balance.Amount {
...
+		}
+	})
+	contributorBalance := balance.NewAmountFromInt(0)
+	for _, fund := range funds {
+		if fund.address.Equal(contributor) {
+			contributorBalance = contributorBalance.Plus(fund.fundingAmount)
+			return contributorBalance, nil
+		}
+	}
+	return nil, ErrWithdrawCheckFundsFailed
```
ws4charlie, line 50: Can we remove this error from return if this func will never encounter any error in its implementation?

client/gov_request_types.go
kerhbal: sure, inside this `GetProposalByIDReply`, I added `ProposalFunds` just to test if my withdraw works.
we can either use this one or come with a new one including proposal funds.

scripts/governance/sdk/actions.py
```diff
@@ -220,3 +278,17 @@ def query_proposal(proposal_id):
     req = {"proposal_id": proposal_id}
     resp = rpc_call('query.GetProposalByID', req)
     return resp["result"]
+
+
+def query_balance(address):
```
ws4charlie, line 283: Are these 2 funcs same?

kerhbal: yea, will delete it... must be a copy paste mistake

action/governance/withdrawFunds.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil {
+		ctx.Logger.Error("No available funds to withdraw for this contributor :", withdrawProposal.Contributor)
+		result := action.Response{
+			Events: action.GetEvent(withdrawProposal.Tags(), "no_available__fund_to_withdraw_for_this_contributor"),
+		}
+		return false, result
+	}
+
+	// 4. Check if the contributor has sufficient funds to withdraw for that proposal
+	withdrawAmount := balance.NewAmountFromBigInt(withdrawProposal.WithdrawValue.Value.BigInt())
```
ws4charlie, line 139: The DeductFunds will do this check inside, right?

action/governance/withdrawFunds.go
```diff
@@ -0,0 +1,234 @@
...
+			Events: action.GetEvent(withdrawProposal.Tags(), "withdraw_proposal_deduct_fund_failed"),
+		}
+		return false, result
+	}
+	// add to beneficiary address
+	coin := withdrawProposal.WithdrawValue.ToCoin(ctx.Currencies)
+	err = ctx.Balances.AddToAddress(withdrawProposal.Beneficiary.Bytes(), coin)
+	if err != nil {
+		// return funds to proposal
+		err = ctx.ProposalMasterStore.ProposalFund.AddFunds(proposal.ProposalID, withdrawProposal.Contributor, withdrawAmount)
```
ws4charlie, line 164: Can you check with Kevin/Tanmay if we need rollback in Tx if something like 'AddToAddress/Minus' this happened? I think this rollback could be already handled somewhere else, because what if this AddFunds rollback fails again and we still lost the money.

action/governance/withdrawFunds.go
```diff
@@ -0,0 +1,234 @@
...
+	}
+
+	result := action.Response{
+		Events: action.GetEvent(withdrawProposal.Tags(), "withdraw_proposal_success"),
+	}
+
+	// check proposal funds:
+	// if all funds for this proposal has been withdrawn, mark outcome cancelled
+	if currentFundsForProposal.BigInt().Cmp(big.NewInt(0)) == 0 {
+		proposal.Outcome = governance.ProposalOutcomeCancelled
```
ws4charlie, line 185: ProposalOutcomeInsufficientFunds in this situation?

ws4charlie: We have a proposal cancel Tx out there.

ws4charlie: And we won't get proposal updated without writing to the store, right?

kerhbal: sorry this one should be deleted, it was my previous design

data/governance/proposal_fund_store.go
```diff
@@ -120,6 +125,19 @@ func (pf *ProposalFundStore) AddFunds(proposalId ProposalID, fundingAddress keys
 	return pf.set(key, *amt.Plus(amount))
 }
 
+func (pf *ProposalFundStore) DeductFunds(proposalId ProposalID, fundingAddress keys.Address, amount *balance.Amount) error {
+	key := storage.StoreKey(string(proposalId) + storage.DB_PREFIX + fundingAddress.String())
```
ws4charlie, line 129: Do we need to delete the record if the amount drops to 0? Confirm with Tanmay?

action/governance/withdrawFunds.go
kerhbal: yeah let me ask. and here I implemented this because there is same logic in createProposal

data/governance/proposal_fund_store.go
kerhbal: sure

action/governance/withdrawFunds.go
```diff
@@ -0,0 +1,234 @@
...
+	if err != nil || !ok {
+		ctx.Logger.Error("Failed to delete proposal from ACTIVE store :", proposal.ProposalID)
+		result := action.Response{
+			Events: action.GetEvent(withdrawProposal.Tags(), "failed_to_delete_proposal_from_active_store"),
+		}
+		return false, result
+	}
+
+	// 3. Check if the contributor has funded this proposal, if so, get the amount of funds
+	proposalFund, err := governance.GetCurrentFundsByContributor(proposal.ProposalID, withdrawProposal.Contributor, ctx.ProposalMasterStore.ProposalFund)
```
ws4charlie, line 129: If all we need to do here is to 'WithdrawFund' and this is the only place func 'GetCurrentFundsByContributor' is used. We actually can move this func's logic into DeductFunds(). This will make the logic here very simple. But it works here and it's up to you :).

kerhbal: yeah, I think put it here will return the error will clearly points out that the contributor has not funded the proposal. and the error message from deduct will only say insufficient funds.

client/gov_request_types.go
ws4charlie: I have a new one for this query. The olclient will allow either listing single proposal by ID or multiple proposals by state.

kerhbal: cool, then we can use yours.

data/governance/proposal_fund_store.go
kerhbal: I think we decide to keep it.

action/governance/withdrawFunds.go
kerhbal: as we spoke, added panic inside

action/governance/withdrawFunds.go
kerhbal: yeah, I'll remove this and test if this is covered

data/governance/proposal_fund.go
kerhbal: discussed, remain no change

PR #524, opened 2020-06-02
OLP-908: End of the Block consolidation for Proposals

KevinOneLedger:
1. Modified Job bus to handle Oneledger Chain jobs.
2. Added new job to check proposal votes.
3. Added an internal transaction to expire votes for a proposal.
4. Refactored Internal Service/Router.

### DIFF
--- a/event/gov_check_votes.go
+++ b/event/gov_check_votes.go
@@ -17,7 +17,7 @@ type JobGovCheckVotes struct {
 	Status     jobs.Status
 }
 
-func NewGovCheckVotes(proposalID governance.ProposalID, status governance.ProposalStatus) *JobGovCheckVotes {
+func NewGovCheckVotesJob(proposalID governance.ProposalID, status governance.ProposalStatus) *JobGovCheckVotes {
 	return &JobGovCheckVotes{
 		ProposalID: proposalID,
 		JobID:      proposalID.String() + storage.DB_PREFIX + strconv.Itoa(int(status)),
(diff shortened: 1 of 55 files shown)


event/gov_check_votes.go
```diff
@@ -0,0 +1,93 @@
...
+	proposal, err := proposalMaster.Proposal.WithPrefixType(governance.ProposalStateActive).Get(j.ProposalID)
+	if err != nil {
+		j.Status = jobs.Failed
+		govCtx.Logger.Error("gov_check_votes: proposal not found in active prefix")
+		return
+	}
+
+	//Check if proposal is in voting state
+	if proposal.Status != governance.ProposalStatusVoting {
+		j.Status = jobs.Failed
```
kingpinXD, line 43: its already checked before creating the job proposal.Status == governance.ProposalStatusVoting  , is this required here ?

event/gov_check_votes.go
```diff
@@ -0,0 +1,93 @@
...
+		return
+	}
+
+	//Create internal transaction and broadcast
+	err = BroadcastGovExpireVotesTx(govCtx, j.ProposalID, j.JobID)
+	if err != nil {
+		j.Status = jobs.Failed
+		govCtx.Logger.Error("gov_check_votes: error broadcasting" + err.Error())
+		return
+	}
```
kingpinXD, line 74: Broadcast might fail becuase a ) TX already present in cache b) Marhlling c) Some other reason ,   Should we panic in cases b and c ?

PR #528, opened 2020-06-03
Domain purchase and reset uri fix

kerhbal:
This fix adds current block height check with config, so that it won't get error when a new node replays the txs.
height to compare is set to 200k

### DIFF
--- a/action/ons/update.go
+++ b/action/ons/update.go
@@ -165,7 +165,7 @@ func runUpdate(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
 			}
 		}
 		d.URI = update.Uri
-	} else if strings.Compare(update.Uri, "") == 0 {
+	} else if ctx.State.Version() >= ctx.Config.ReleaseChanges.DomainChange1 && strings.Compare(update.Uri, "") == 0 {
 		// reset uri
 		d.URI = ""
 	}
--- a/config/server.go
+++ b/config/server.go
@@ -452,12 +452,12 @@ type EthereumChainDriverConfig struct {
 }
 
 type ReleaseChangesConfig struct {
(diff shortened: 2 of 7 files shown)


action/ons/update.go
```diff
@@ -165,7 +165,7 @@ func runUpdate(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
 			}
 		}
 		d.URI = update.Uri
-	} else if strings.Compare(update.Uri, "") == 0 {
+	} else if ctx.State.Version() >= ctx.Config.ReleaseChanges.DomainChange1 && strings.Compare(update.Uri, "") == 0 {
```
pav-kap, line 168: Are we checking for the condition that "ctx.Config.ReleaseChanges.DomainChange1" doesn't exist in config.toml?

config/server.go
```diff
@@ -449,6 +451,17 @@ type EthereumChainDriverConfig struct {
 	Connection string `toml:"connection" desc:"ethereum node connection url default: http://localhost:7545"`
 }
 
+type ReleaseChangesConfig struct {
+	DomainChange1 int64 `toml:"domainChange_1" desc:"changes related to purchase and update domain"`
+}
+
+func DefaultReleaseChangesConfig() *ReleaseChangesConfig {
+	var cfg ReleaseChangesConfig
+	cfg.DomainChange1 = 200000
```
pav-kap, line 460: can we have it config.toml as 200000 and here it can default to 0 right?

action/ons/update.go
kerhbal: no, this ReleaseChanges.DomainChange1 is defined in a function, and config.toml is just a place to export those configs

config/server.go
alexjlan: I think we can make 200k here, but change the devnet command to have it become 0 so that our test scripts works

PR #531, opened 2020-06-05
Cancel proposal

ws4charlie:
Implementation of Tx calcel proposal
https://oneledger.atlassian.net/browse/OLP-903


action/governance/cancelProposal.go
```diff
@@ -0,0 +1,166 @@
...
+	err = pms.Proposal.WithPrefixType(gov.ProposalStateFailed).Set(proposal)
+	if err != nil {
+		return false, action.Response{
+			Log: fmt.Sprintf("cancel proposal failed, id= %v, failed to add proposal to FAILED store", cc.ProposalId)}
+	}
+
+	// Delete proposal in ACTIVE store
+	ok, err := pms.Proposal.WithPrefixType(gov.ProposalStateActive).Delete(cc.ProposalId)
+	if err != nil || !ok {
+		return false, action.Response{
```
KevinOneLedger, line 161: Might have to panic if it failed to delete.

ws4charlie: Yes. You're right. I wanted to add this panic until I saw that the Delete always returns a Nil.

PR #533, opened 2020-06-08
Olp 981

kingpinXD:
Added Finalization to Blockender
Changed finalization from Status to State


kingpinXD:
Yeah removed the binary ,and that is a separate task

PR #535, opened 2020-06-09
Query by proposal ID, type and proposer

ws4charlie:
https://oneledger.atlassian.net/browse/OLP-909
https://oneledger.atlassian.net/browse/OLP-911
https://oneledger.atlassian.net/browse/OLP-912

### DIFF
--- a/cmd/olclient/gov.go
+++ b/cmd/olclient/gov.go
@@ -35,6 +35,8 @@ type VoteArguments struct {
 type ListProposalArguments struct {
 	ProposalID string `json:"proposalID"`
 	State      string `json:"state"`
+	Proposer   []byte `json:"proposer"`
+	Type       string `json:"type"`
 }
 
 var (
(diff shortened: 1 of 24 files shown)


cmd/olclient/gov.go
```diff
@@ -191,38 +195,49 @@ func listProposals(cmd *cobra.Command, args []string) error {
...
+		req := client.ListProposalRequest{
+			ProposalId: governance.ProposalID(listArgs.ProposalID),
+		}
+
+		reply, err := fullnode.ListProposal(req)
+		if err != nil {
+			return errors.New("error in getting proposal")
+		}
+
+		ps := reply.ProposalStats[0]
```
KevinOneLedger, line 209: You might want to create another struct for the single proposal reply. It looks strange when you have to reference the first element.

scripts/governance/expireVotes.py
```diff
@@ -47,23 +47,23 @@ def completed_votes():
 
     # 2nd fund
     fund_proposal(encoded_pid, _each_funding, addr_list[1])
```
KevinOneLedger, line 49: Why do we check the outcome instead of the state? It won't change until it's completed. Not a big deal, just a question.

service/query/governance.go
```diff
@@ -4,87 +4,94 @@ import (
...
+		Funds:    *funds,
+		Votes:    *stat,
 	}
-
 	*reply = client.ListProposalsReply{
-		Proposals:     proposals,
-		ProposalFunds: proposalFunds,
-		ProposalVotes: proposalVotes,
-		State:         proposalState,
+		ProposalStats: []client.ProposalStat{ps},
```
KevinOneLedger, line 30: Mentioned this above, you can change the response type to use a single ProposalStat.

cmd/olclient/gov.go
ws4charlie: That was exactly I had before. From SDK side, Freddy said they prefer the unified way, like what we did in ONS queries.

service/query/governance.go
ws4charlie: Same as above.

PR #558, opened 2020-06-12
Changed proposal query parameter STATE/TYPE from string to integer and misc changes

ws4charlie:


cmd/olclient/gov.go
```diff
@@ -197,10 +197,16 @@ func listProposals(cmd *cobra.Command, args []string) error {
 
 	// List single proposal
 	if listArgs.ProposalID != "" {
-		req := client.ListProposalRequest{
-			ProposalId: governance.ProposalID(listArgs.ProposalID),
+		pID := governance.ProposalID(listArgs.ProposalID)
+		if listArgs.ProposalID != "" {
```
kerhbal, line 201: why there is duplicate of `if listArgs.ProposalID != ""`?
and inside pID.Err() we are still checking if proposal id is empty

cmd/olclient/gov.go
```diff
@@ -210,10 +216,28 @@ func listProposals(cmd *cobra.Command, args []string) error {
 		printProposal(ps.Proposal, ps.Funds, &ps.Votes)
 		fmt.Println("Height: ", reply.Height)
 	} else { // List multiple proposals
+		pState := governance.NewProposalState(listArgs.State)
+		pType := governance.NewProposalType(listArgs.Type)
+		proposer := keys.Address(listArgs.Proposer)
+		if listArgs.State != "" {
+			if pState == governance.ProposalStateInvalid {
```
kerhbal, line 223: what if the pState here is not valid and not equal to ProposalStateInvalid(0xEE)?
similar for pType

cmd/olclient/gov.go
ws4charlie: > why there is duplicate of `if listArgs.ProposalID != ""`? (line 199 and line 201)
> and inside pID.Err() we are still checking if proposal id is empty

Correct. I'll remove this unnecessary check.

cmd/olclient/gov.go
ws4charlie: > what if the pState here is not valid and not equal to ProposalStateInvalid(0xEE)?
> similar for pType

If pState or pType is not a valid one, then we return error. Something not correct here?

PR #570, opened 2020-06-18
OLP-1017: Rewards DB creation and wrapper methods to interact

KevinOneLedger:


data/rewards/store.go
```diff
@@ -0,0 +1,95 @@
...
+func NewRewardStore(prefix string, state *storage.State) *RewardStore {
+	return &RewardStore{
+		State:  state,
+		szlr:   serialize.GetSerializer(serialize.PERSISTENT),
+		prefix: storage.Prefix(prefix),
+	}
+}
+
+func (rs *RewardStore) WithState(state *storage.State) {
+	rs.State = state
```
kingpinXD, line 28: You might want to  return the Store

data/rewards/store.go
```diff
@@ -0,0 +1,95 @@
...
+	"github.com/Oneledger/protocol/storage"
+	"strings"
+)
+
+type RewardStore struct {
+	State  *storage.State
+	szlr   serialize.Serializer
+	prefix []byte
+
+	rewardOptions Options
```
kingpinXD, line 16: We should use a pointer to the store instead , just to follow the convention we already have

PR #573, opened 2020-06-19
Olp 1012 fund pool

kingpinXD:


service/tx/service.go
```diff
@@ -98,6 +98,53 @@ func (svc *Service) SendTx(args client.SendTxRequest, reply *client.CreateTxRepl
 	return nil
 }
 
+func (svc *Service) SendPoolTx(args client.SendPoolTxRequest, reply *client.CreateTxReply) error {
```
KevinOneLedger, line 101: Is this function only for testing? When you try to sign with the from address, the node wallet might not have the private key.

service/tx/service.go
```diff
@@ -98,6 +98,53 @@ func (svc *Service) SendTx(args client.SendTxRequest, reply *client.CreateTxRepl
...
+		Data: data,
+		Fee:  fee,
+		Memo: uuidNew.String(),
+	}
+
+	if _, err := svc.accounts.GetAccount(args.From); err != nil {
+		return accounts.ErrGetAccountByAddress
+	}
+
+	pubKey, signed, err := svc.accounts.SignWithAddress(tx.RawBytes(), send.From)
```
KevinOneLedger, line 126: If the account is not in the node wallet, this might fail.

PR #584, opened 2020-07-02
Internal tx olp 1045

kingpinXD:
Added Mechanism to store and execute Internal transactions


app/controller.go
```diff
@@ -271,9 +270,12 @@ func (app *App) blockEnder() blockEnder {
...
 		doEthTransitions(app.Context.jobStore, app.Context.ethTrackers, app.Context.node.ValidatorAddress(), ethTrackerlog, app.Context.witnesses, app.Context.deliver)
 
-		//Check for vote expiration on active proposals
-		updateProposals(app.Context.proposalMaster, app.Context.jobStore, app.Context.deliver)
+		//updateProposals(app.Context.proposalMaster, app.Context.jobStore, app.Context.deliver)
+		ExpireProposals(&app.header, &app.Context, app.logger)
+
+		FinalizeProposals(&app.header, &app.Context, app.logger)
 
+		//Check for vote expiration on active proposals
```
KevinOneLedger, line 278: You can clean up some of the comments, looks like it is out of order.

app/internalTX.go
```diff
@@ -0,0 +1,197 @@
...
+		rawTx := action.RawTx{
+			Type: action.PROPOSAL_FINALIZE,
+			Data: txData,
+			Fee:  action.Fee{},
+			Memo: uuidNew.String(),
+		}
+		ok, _ := newFinalize.ProcessDeliver(actionctx, rawTx)
+		if !ok {
+			logger.Error("Failed to Finalize : ", txData, "Error : ", err)
+			return
```
KevinOneLedger, line 143: You think it might be better to break here? So we can clear the db with the delete

kingpinXD: I gues the delete , deletes everything ,irrespective of whether deliver was a success or a failure ,but in case it was not a success , it would still be in the passed /failed state , and be picked up again in the next block. 
not breaking would allow us to process the other tx's in the queue

PR #585, opened 2020-07-03
OLP-1015:Block Ender Logic 2: Functionality for every 150,000 blocks

KevinOneLedger:


app/controller.go
```diff
@@ -480,18 +461,35 @@ func handleBlockRewards(validators *identity.ValidatorStore, rewards *rewards.Re
 			continue
```
kingpinXD, line 461: If the Validator votes in the last block , and is removed from the validator list in the present block , would he still get the rewards for his last vote ?

app/controller.go
```diff
@@ -456,22 +450,9 @@ func handleBlockRewards(validators *identity.ValidatorStore, rewards *rewards.Re
...
-		}
-		if len(validatorVote.Signature) == 0 {
-			continue
-		}
-
+	for _, vote := range votes {
 		//Verify Validator Address
-		valAddress := keys.Address(validatorVote.ValidatorAddress)
+		valAddress := keys.Address(vote.Validator.Address)
 		if valAddress.Err() != nil {
```
kingpinXD, line 456: We get the Validator address from tendermint , wondering if this error can ever happen ?

kingpinXD, line 456: We get the address from Tendermint , wondering if this error can ever happen ?

app/controller.go
KevinOneLedger: Yea, technically the validator participated in creating the block so it will get the reward for that one.

app/controller.go
KevinOneLedger: Its more like a validation after casting

app/controller.go
kingpinXD: The staking updates in the blockender of the previous block, I assumed that the present block would have the updated validator list.And Get would iterate over that. 
But as long as it's tested, I am fine with it

PR #586, opened 2020-07-03
OLP-1034: fixed self-staking issues: the crash; incremental validator update; stop staking from multiple addresses

ws4charlie:
https://oneledger.atlassian.net/browse/OLP-1034

### DIFF
--- a/action/staking/stake.go
+++ b/action/staking/stake.go
@@ -3,12 +3,13 @@ package staking
 import (
 	"encoding/json"
 
+	"github.com/pkg/errors"
+	"github.com/tendermint/tendermint/libs/kv"
+
 	"github.com/Oneledger/protocol/action"
 	"github.com/Oneledger/protocol/data/balance"
 	"github.com/Oneledger/protocol/data/keys"
 	"github.com/Oneledger/protocol/identity"
-	"github.com/pkg/errors"
-	"github.com/tendermint/tendermint/libs/kv"
 )
 
 var _ action.Msg = &Stake{}
@@ -137,7 +138,7 @@ func (s stakeTx) ProcessFee(ctx *action.Context, signedTx action.SignedTx, start
 
 // allow validator's stake address update only when it's clean
 func isStakeAddressClean(ctx *action.Context, v *identity.Validator) (bool, error) {
-	options, err := ctx.Govern.GetStakingOptions()
+	options, err := ctx.GovernanceStore.GetStakingOptions()
 	if err != nil {
 		return false, err
 	}
--- a/action/staking/stake.go
+++ b/action/staking/stake.go
@@ -4,6 +4,7 @@ import (
 	"encoding/json"
 
 	"github.com/Oneledger/protocol/action"
+	"github.com/Oneledger/protocol/data/balance"
 	"github.com/Oneledger/protocol/data/keys"
 	"github.com/Oneledger/protocol/identity"
 	"github.com/pkg/errors"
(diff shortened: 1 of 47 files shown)


action/staking/stake.go
```diff
@@ -134,6 +136,41 @@ func (s stakeTx) ProcessFee(ctx *action.Context, signedTx action.SignedTx, start
...
+	}
+
+	// check pending amount
+	pendingAmounts := ctx.Delegators.GetMaturedPendingAmount(v.StakeAddress, ctx.Header.Height, options.MaturityTime+1)
+	if len(pendingAmounts) != 0 {
+		return false, nil
+	}
+
+	// check bounded amount
+	boundCoin, err := ctx.Delegators.GetDelegatorBoundedAmount(v.StakeAddress)
```
KevinOneLedger, line 163: What does the bounded amount mean?

identity/validator_set.go
```diff
@@ -310,20 +362,25 @@ func (vs *ValidatorStore) GetEndBlockUpdate(ctx *ValidatorContext, req types.Req
 	minSelfDelegationAmount := stakingOptions.MinSelfDelegationAmount.BigInt().Int64()
 
 	if height > 1 || (len(vs.byzantine) > 0) {
+		// map for no top-power validators
```
KevinOneLedger, line 365: You mean non top-power validators?

identity/validator_set.go
```diff
@@ -366,6 +425,39 @@ func (vs *ValidatorStore) GetEndBlockUpdate(ctx *ValidatorContext, req types.Req
...
+			if !ok {
+				continue
+			}
+			// get last purge height
+			purgeHeight, err := vs.GetLastPurgeHeight(keys.Address(addr))
+			if err != nil {
+				logger.Errorf("failed to get last purge height, validator: %s", addrHuman)
+				continue
+			}
+			// validator purged in block H persists in block H+1, H+2, so we can't purge it again
```
KevinOneLedger, line 442: So basically we don't need to purge 2 blocks after the original purge since the validator will persist in the list for those blocks. This means that it takes 2 blocks for the purge to take effect.

identity/validator_set.go
```diff
@@ -366,6 +425,39 @@ func (vs *ValidatorStore) GetEndBlockUpdate(ctx *ValidatorContext, req types.Req
 				}
 			}
 		}
+
+		// purge all other active validators if not among the top
+		for addr, power := range vs.lastActive {
+			addrHuman := keys.Address(addr).Humanize()
+			pub, ok := noTopValidators[addrHuman]
```
KevinOneLedger, line 432: You're just checking the NON validator map for the address to determine if it needs to be purged.

action/staking/stake.go
ws4charlie: @KevinOneLedger I was initially confused by the word 'bounded'. I figured out from his implementation that the word 'bounded' here is the same as 'withdrable' amount.

identity/validator_set.go
ws4charlie: > You mean non top-power validators?

Yeah. That's the correct word and I'll correct it.

identity/validator_set.go
ws4charlie: > So basically we don't need to purge 2 blocks after the original purge since the validator will persist in the list for those blocks. This means that it takes 2 blocks for the purge to take effect.

Yes. The tendermint doc says that and it actually does that as well which crashed my node again in my first fix.

PR #588, opened 2020-07-07
Withdraw tx olp 1016

kingpinXD:


action/rewards/withdraw.go
```diff
@@ -0,0 +1,133 @@
...
+
+	"github.com/Oneledger/protocol/action"
+	"github.com/Oneledger/protocol/action/helpers"
+	"github.com/Oneledger/protocol/data/balance"
+	"github.com/Oneledger/protocol/data/rewards"
+)
+
+type Withdraw struct {
+	ValidatorAddress        action.Address `json:"validatorAddress"`
+	ValidatorSigningAddress action.Address `json:"validatorSigningAddress"`
```
KevinOneLedger, line 17: If we use a separate signing address, how can we check if it is owned by the validator?

action/rewards/withdraw.go
```diff
@@ -0,0 +1,133 @@
...
+		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrUnserializable, withdraw.Tags(), err)
+	}
+
+	//2. Get the difference of amount earned vs amount withdrawn for the validator issuing this transaction
+	//3. Check how much he is eligible to withdraw
+	//4. If the amount withdrawn is less than or equal to amount eligible to be withdrawn, make the transaction success.
+	//5. In case of no failure, add this amount the person withdrew, to total withdrawn amount in cumulative rewards db
+	//maturedBalance, _ := ctx.RewardMasterStore.RewardCm.GetMaturedRewards(withdraw.ValidatorAddress)
+	//fmt.Println("Matured Balance :", maturedBalance, withdraw.ValidatorAddress)
+	err = ctx.RewardMasterStore.RewardCm.WithdrawRewards(withdraw.ValidatorAddress, &withdraw.WithdrawAmount.Value)
```
KevinOneLedger, line 120: I See that you're only withdrawing from the rewards store. What about the reward Pool?

action/rewards/withdraw.go
kingpinXD: Yeah , the assumption was ,if someone else wants to pay for a trasfer of money to the validator ,he can ,the funds still go to the validator address ,but you are right , I think ill do this through olclient  ,That would make it more secure

PR #591, opened 2020-07-08
Fix one address for stake

DeV1doR:


action/staking/stake.go
```diff
@@ -92,6 +92,11 @@ func (s stakeTx) Validate(ctx *action.Context, tx action.SignedTx) (bool, error)
 		return false, err
 	}
 
+	val, err := ctx.Validators.Get(st.ValidatorAddress)
```
alexjlan, line 95: in the validate function, we should not read from store, validate only validate tx by the data itself

action/staking/unstake.go
```diff
@@ -88,6 +88,11 @@ func (us unstakeTx) Validate(ctx *action.Context, tx action.SignedTx) (bool, err
 		return false, err
 	}
 
+	val, err := ctx.Validators.Get(ust.ValidatorAddress)
```
alexjlan, line 91: same as stake

action/staking/withdraw.go
```diff
@@ -90,6 +90,11 @@ func (withdrawTx) Validate(ctx *action.Context, tx action.SignedTx) (bool, error
 		return false, err
 	}
 
+	val, err := ctx.Validators.Get(draw.ValidatorAddress)
```
alexjlan, line 93: same as above

PR #594, opened 2020-07-09
Olp 1051 Query for Withdrawal Amount by Proposal Id and Account Address

kerhbal:


service/query/governance.go
```diff
@@ -83,6 +83,23 @@ func (svc *Service) ListProposals(req client.ListProposalsRequest, reply *client
 	return nil
 }
 
+// list funds by funder for a proposal
+func (svc *Service) GetFundsForProposalByFunder(req client.GetFundsForProposalByFunderRequest, reply *client.GetFundsForProposalByFunderReply) error {
+	// Validate parameters
+	if len(req.Funder) != 0 {
```
ws4charlie, line 89: req.Funder.Err() has already checked zero length error inside I think.

kerhbal: yeah, will remove if here

PR #598, opened 2020-07-13
Olp 1035 Self Staking Integration Tests

kerhbal:


scripts/staking/self_staking.py
```diff
@@ -0,0 +1,34 @@
...
+
+_wait = 20
+
+if __name__ == "__main__":
+    staking_test = Staking(node_4)
+    staking_test.prepare()
+    addr_list = addresses()
+
+    staking_test.stake('3', addr_list[1])
+    staking_test.checkStatus(addr_list[1], 3, 0, False)
```
ws4charlie, line 13: Can you add staking/unstake from a different address(like addr_list[2]) after your 1st staking? They should fail(due to stake address mismatches).

kerhbal: yeah sure

PR #602, opened 2020-07-15
Validations olp 1041

kingpinXD:


data/governance/validations.go
```diff
@@ -0,0 +1,333 @@
...
+	if oldOptions.BlockConfirmation != opt.BlockConfirmation {
+		return false, errors.New("BlockConfirmations cannot be changed")
+	}
+	//if !verifyRangeInt64(opt.BlockConfirmation, minBlockConfirmation, maxBlockConfirmation) {
+	//	return false, errors.New("BlockConfirmations should be between 0 and 50")
+	//}
+	return true, nil
+}
+
+func (st *Store) ValidateBTC(opt *bitcoin.ChainDriverOption) (bool, error) {
```
alexjlan, line 188: since we are not supporting BTC, I think we should just make this unchangeable.

data/governance/validations.go
```diff
@@ -0,0 +1,333 @@
...
+	}
+	ok, err = verifyDistribution(general.PassedFundDistribution)
+	if err != nil || !ok {
+		return false, err
+	}
+	ok, err = verifyDistribution(general.FailedFundDistribution)
+	if err != nil || !ok {
+		return false, err
+	}
+	if code.ProposalExecutionCost != oldOptions.CodeChange.ProposalExecutionCost {
```
alexjlan, line 279: "me" seems to be a typo

alexjlan: and all those sentence doesn't make sense

data/governance/validations.go
```diff
@@ -0,0 +1,333 @@
...
+	if general.ProposalExecutionCost != oldOptions.General.ProposalExecutionCost {
+		return false, errors.New("OnLedger Address cannot me changed general update execution cost")
+	}
+	if opt.BountyProgramAddr != oldOptions.BountyProgramAddr {
+		return false, errors.New("OnLedger Address cannot me changed Bounty Program address")
+	}
+	return true, nil
+}
+
+func (st *Store) ValidateRewards(opt *rewards.Options) (bool, error) {
```
alexjlan, line 294: wait a second, block reword can't be changed?

data/governance/validations.go
kingpinXD: You mean the error statements ? I just added Oneledger Address ,to signify that it is something that is set by us.I can change it to something else ,do you mean like " Bounty Address cannot be changed " ,

data/governance/validations.go
kingpinXD: The update TX has  all the fields , I just verify whether the new fields are exactly equal to old or not. So it is unchangeable. If we want to remove this check altogether , we would need to fetch the BTC options from a separate place and not gov options ,is that what you are referring to ?

alexjlan: it's okay as it right now, I got it

data/governance/validations.go
alexjlan: it's better now.

PR #603, opened 2020-07-16
Block rewards calculate logics and unit tests cases

ws4charlie:
1. the calculation logics
2 the test cases covered simulated 5-years distribution scenario.


data/rewards/init.go
```diff
@@ -24,6 +35,17 @@ type RewardMasterStore struct {
 	RewardCm *RewardCumulativeStore
 }
 
+type RewardMasterState struct {
```
KevinOneLedger, line 38: I was a little confused why we use a separate struct to dump the state. I'm guessing this is to wrap any other data we want to dump in the future?

go.mod
```diff
@@ -8,17 +8,15 @@ require (
...
 	github.com/deckarep/golang-set v1.7.1 // indirect
 	github.com/elastic/gosigar v0.10.5 // indirect
 	github.com/ethereum/go-ethereum v1.9.6
 	github.com/fjl/memsize v0.0.0-20190710130421-bcb5799ab5e5 // indirect
 	github.com/gballet/go-libpcsclite v0.0.0-20190607065134-2772fd86a8ff // indirect
-	github.com/go-delve/delve v1.4.1 // indirect
 	github.com/go-kit/kit v0.10.0
-	github.com/go-language-server/uri v0.2.0 // indirect
 	github.com/google/uuid v1.1.1
+	github.com/hashicorp/golang-lru v0.5.4 // indirect
```
KevinOneLedger, line 19: Double check your dependencies, not sure why they're different.

app/application.go
```diff
@@ -165,7 +165,6 @@ func (app *App) setupState(stateBytes []byte) error {
 		return errors.Wrap(err, "Setup FeeOptions Options")
 	}
 	app.Context.feePool.SetupOpt(&initial.Governance.FeeOption)
-	app.Context.rewardMaster.SetOptions(&initial.Governance.RewardOptions)
```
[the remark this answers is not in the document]
ws4charlie: @KevinOneLedger We already set options on line 153. This line is redundant.

app/controller.go
```diff
@@ -425,9 +427,20 @@ func doEthTransitions(js *jobs.JobStore, ts *ethereum.TrackerStore, myValAddr ke
 
 }
 
-func handleBlockRewards(validators *identity.ValidatorStore, rewardMaster *rewards.RewardMasterStore, block RequestBeginBlock) abciTypes.Event {
+//Get Individual Validator reward based on power
+func getRewardForValidator(totalPower int64, validatorPower int64, totalRewards *balance.Amount) *balance.Amount {
+	percent := int64(float64(validatorPower) / float64(totalPower) * 1000.0)
```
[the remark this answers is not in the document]
ws4charlie: @KevinOneLedger Good catch. I'll change.

data/rewards/init.go
ws4charlie: Yep. This is to wrap any other data from the stores we want to dump in the future, e.g. pending rewards.

go.mod
ws4charlie: I used 'go mod tidy' and it ended up like this. Is 'go mod tidy' safe?

ws4charlie: Let me roll back the changes to play safe.

PR #606, opened 2020-07-22
added public key to result of show_node_id command

ws4charlie:
1. We found a way to create a secure wallet account associated with genesis stake address.
2. Added public key information to 'olclient show_node_id' and 'olfullnode show_node_id', so node operator is able to create wallet account by specifying public key and their private key.

### DIFF
--- a/client/request_types.go
+++ b/client/request_types.go
@@ -141,7 +141,7 @@ type NodeIDRequest struct {
 	ShouldShowIP bool `json:"shouldShowIP,omitempty"`
 }
 type NodeIDReply struct {
-	PublicKey string `json:publicKey`
+	PublicKey string `json:"publicKey"`
 	Id        string `json:"id"`
 }
 
--- a/client/request_types.go
+++ b/client/request_types.go
@@ -141,7 +141,8 @@ type NodeIDRequest struct {
 	ShouldShowIP bool `json:"shouldShowIP,omitempty"`
 }
 type NodeIDReply struct {
-	Id string `json:"id"`
+	PublicKey string `json:publicKey`
+	Id        string `json:"id"`
 }
 
 type AddAccountRequest = accounts.Account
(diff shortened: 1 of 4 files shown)


client/request_types.go
```diff
@@ -141,7 +141,8 @@ type NodeIDRequest struct {
 	ShouldShowIP bool `json:"shouldShowIP,omitempty"`
 }
 type NodeIDReply struct {
-	Id string `json:"id"`
+	PublicKey string `json:publicKey`
```
kerhbal, line 144: do you mean \`json:"publicKey"\` ?

cmd/olfullnode/show_node_id.go
```diff
@@ -64,6 +66,12 @@ func showNodeID(root string, shouldShowIP bool) (result string, err error) {
 		}
 		return fmt.Sprintf("%s@%s", nodeKey.ID(), u.Host), nil
 	} else {
-		return string(nodeKey.ID()), nil
+		pubKey, _ := keys.PubKeyFromTendermint(nodeKey.PubKey().Bytes())
```
kerhbal, line 69: why this logic only exists in the scenario that we don't show IP? is this intentional?

service/node/service.go
```diff
@@ -61,7 +63,14 @@ func (svc *Service) ID(req client.NodeIDRequest, reply *client.NodeIDReply) erro
 		out := fmt.Sprintf("%s@%s", nodeKey.ID(), u.Host)
 		*reply = client.NodeIDReply{Id: out}
 	} else {
-		*reply = client.NodeIDReply{Id: string(nodeKey.ID())}
+		pubKey, _ := keys.PubKeyFromTendermint(nodeKey.PubKey().Bytes())
```
kerhbal, line 66: same question as above

PR #610, opened 2020-07-27
Fixed block rewards calculation in catching up scenario

ws4charlie:
The issue was caused by a different rewards amount from calculation when node is catching up blocks. A block's rewards amount is based on 2 factors, the block generation speed and the amount we have already distributed so far. The difference came from the second factor, the amount of rewards we have already distributed, in catch-up scenario. Now we persisted that amount in DB so we can get the same rewards amount even in catch-up mode.

### DIFF
--- a/data/rewards/calculator.go
+++ b/data/rewards/calculator.go
@@ -9,9 +9,10 @@ import (
 )
 
 type RewardYear struct {
-	StartTime   time.Time
-	CloseTime   time.Time
-	Distributed *balance.Amount
+	StartTime     time.Time
+	CloseTime     time.Time
+	Distributed   *balance.Amount
+	TillLastCycle *balance.Amount // distributed till last cycle
 }
 
 type RewardYears struct {
@@ -33,6 +34,11 @@ func NewRewardCached() RewardCached {
 		burnedout: false,
 		amount:    balance.NewAmount(0),
 	}
+}
+
+func (cache *RewardCached) available() bool {
+	// we can always recalculate. This is just for performance purpose.
+	return cache.cycleNo > 0
 }
 
 type RewardCalculator struct {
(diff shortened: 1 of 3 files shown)


data/rewards/calculator.go
```diff
@@ -9,9 +9,10 @@ import (
 )
 
 type RewardYear struct {
-	StartTime   time.Time
-	CloseTime   time.Time
-	Distributed *balance.Amount
+	StartTime     time.Time
+	CloseTime     time.Time
+	Distributed   *balance.Amount
+	TillLastCycle *balance.Amount // distributed till last cycle
```
KevinOneLedger, line 15: Do you mean the amount that has been distributed up to the last cycle? Not including the current cycle?

data/rewards/calculator.go
```diff
@@ -84,10 +93,10 @@ func (calc *RewardCalculator) Calculate() (amt *balance.Amount, err error) {
 
 	// how much rewards left before year close
 	yearSupply := options.YearBlockRewardShares[year]
-	yearDistributed := calc.rewardYears.Years[year].Distributed
+	yearDistributed := calc.rewardYears.Years[year].TillLastCycle
```
KevinOneLedger, line 96: To clarify, you're persisting the amount accumulated in the last cycle to separate it from the current distribution. So basically the amount left for the year doesn't take into account the amount Distributed for the current cycle?

data/rewards/calculator.go
ws4charlie: > Do you mean the amount that has been distributed up to the last cycle? Not including the current cycle?

Yes. Exactly. This is the amount I should use to forecast the rewards for next cycle.

data/rewards/calculator.go
ws4charlie: > To clarify, you're persisting the amount accumulated in the last cycle to separate it from the current distribution. So basically the amount left for the year doesn't take into account the amount Distributed for the current cycle?

That's exactly what I am doing here.

PR #615, opened 2020-07-28
Olp 1079 secure wallet

kingpinXD:


service/tx/service.go
```diff
@@ -157,23 +148,22 @@ func (svc *Service) SendPoolTx(args client.SendPoolTxRequest, reply *client.Crea
 	return nil
 }
 
-func (svc *Service) CreateRawSend(args client.SendTxRequest, reply *client.CreateTxReply) error {
```
KevinOneLedger, line 160: Did you remove this CreateRawSend function completely?

KevinOneLedger: Never mind I just missed the function.

PR #618, opened 2020-08-06
Fix block reward CI failure

ws4charlie:


ws4charlie:
@kingpinXD Do you have time to take a look?

Makefile
```diff
@@ -111,6 +111,7 @@ transfertest: reset
 rewardtest: reset
 	@./scripts/testsend
 	python scripts/reward/testRewards.py
+	make reset
```
kingpinXD, line 114: do we need this becuase testRewards also creates wallet addresses ?

ws4charlie: > do we need this becuase testRewards also creates wallet addresses ?

Yes. Both create wallet addresses. Without 'reset' the 2nd test somehow doesn't work well on the \keystore\ left by 1st test.

kingpinXD: Yeah i noticed that as well, if we use the same pub /private key pair to generate two secure walllet accounts they stop working, 
Maybe we should look at it in a different ticket

ws4charlie: Yep. Can you approve this one and let Mani merge CI first

PR #631, opened 2020-08-20
Penalizing

ws4charlie:
For penalizing options review

### DIFF
--- a/action/errors.go
+++ b/action/errors.go
@@ -18,6 +18,9 @@ var (
 	ErrNotEnoughFund      = codes.ProtocolError{codes.TxErrInsufficientFunds, "not enough fund"}
 	ErrGasOverflow        = codes.ProtocolError{codes.TxErrGasOverflow, "gas used exceed limit"}
 	ErrInvalidExtTx       = codes.ProtocolError{codes.TxErrInvalidExtTx, "invalid external tx"}
+	ErrFrozenValidator    = codes.ProtocolError{codes.TxErrInvalidExtTx, "frozen validator"}
+	ErrNonFrozenValidator = codes.ProtocolError{codes.TxErrInvalidExtTx, "non frozen validator"}
+	ErrNonActiveValidator = codes.ProtocolError{codes.TxErrInvalidPubKey, "non active validator"}
 
 	ErrInvalidAddress          = codes.ErrBadAddress
 	ErrInvalidCurrency         = codes.ProtocolError{codes.TxErrInvalidFeeCurrency, "invalid amount"}
--- a/action/errors.go
+++ b/action/errors.go
@@ -18,9 +18,9 @@ var (
 	ErrNotEnoughFund      = codes.ProtocolError{codes.TxErrInsufficientFunds, "not enough fund"}
 	ErrGasOverflow        = codes.ProtocolError{codes.TxErrGasOverflow, "gas used exceed limit"}
 	ErrInvalidExtTx       = codes.ProtocolError{codes.TxErrInvalidExtTx, "invalid external tx"}
-	ErrFrozenValidator    = codes.ProtocolError{codes.TxErrInvalidExtTx, "frozen validator"}
-	ErrNonFrozenValidator = codes.ProtocolError{codes.TxErrInvalidExtTx, "non frozen validator"}
-	ErrNonActiveValidator = codes.ProtocolError{codes.TxErrInvalidPubKey, "non active validator"}
+	ErrFrozenValidator    = codes.ProtocolError{codes.TxErrMaliciousValidator, "frozen validator"}
+	ErrNonFrozenValidator = codes.ProtocolError{codes.TxErrMaliciousValidator, "non frozen validator"}
+	ErrNonActiveValidator = codes.ProtocolError{codes.TxErrMaliciousValidator, "non active validator"}
 
 	ErrInvalidAddress          = codes.ErrBadAddress
 	ErrInvalidCurrency         = codes.ProtocolError{codes.TxErrInvalidFeeCurrency, "invalid amount"}
--- a/action/errors.go
+++ b/action/errors.go
@@ -18,9 +18,6 @@ var (
 	ErrNotEnoughFund      = codes.ProtocolError{codes.TxErrInsufficientFunds, "not enough fund"}
 	ErrGasOverflow        = codes.ProtocolError{codes.TxErrGasOverflow, "gas used exceed limit"}
 	ErrInvalidExtTx       = codes.ProtocolError{codes.TxErrInvalidExtTx, "invalid external tx"}
-	ErrFrozenValidator    = codes.ProtocolError{codes.TxErrMaliciousValidator, "frozen validator"}
-	ErrNonFrozenValidator = codes.ProtocolError{codes.TxErrMaliciousValidator, "non frozen validator"}
-	ErrNonActiveValidator = codes.ProtocolError{codes.TxErrMaliciousValidator, "non active validator"}
 
 	ErrInvalidAddress          = codes.ErrBadAddress
 	ErrInvalidCurrency         = codes.ProtocolError{codes.TxErrInvalidFeeCurrency, "invalid amount"}
--- a/action/penalization/allegation.go
+++ b/action/penalization/allegation.go
@@ -1,0 +1,133 @@
+package penalization
+
+import (
+	"encoding/json"
+
+	"github.com/pkg/errors"
+	"github.com/tendermint/tendermint/libs/kv"
+
+	"github.com/Oneledger/protocol/action"
+	gov "github.com/Oneledger/protocol/data/governance"
+	"github.com/Oneledger/protocol/data/keys"
+)
+
+var _ action.Msg = &Allegation{}
+
+type Allegation struct {
+	ValidatorAddress keys.Address
+	MaliciousAddress keys.Address
+	BlockHeight      int64
+	ProofMsg         string
+}
+
+func (r Allegation) Marshal() ([]byte, error) {
+	return json.Marshal(r)
+}
+
+func (r *Allegation) Unmarshal(data []byte) error {
+	return json.Unmarshal(data, r)
+}
+
+func (r Allegation) Signers() []action.Address {
+	return []action.Address{r.ValidatorAddress.Bytes()}
+}
+
+func (r Allegation) Type() action.Type {
+	return action.ALLEGATION
+}
+
+func (r Allegation) Tags() kv.Pairs {
+	tags := make([]kv.Pair, 0)
+
+	tag := kv.Pair{
+		Key:   []byte("tx.type"),
+		Value: []byte(r.Type().String()),
+	}
+	tag2 := kv.Pair{
+		Key:   []byte("tx.validator"),
+		Value: r.ValidatorAddress.Bytes(),
+	}
+
+	tags = append(tags, tag, tag2)
+	return tags
+}
+
+var _ action.Tx = allegationTx{}
+
+type allegationTx struct{}
+
+func (atx allegationTx) Validate(ctx *action.Context, tx action.SignedTx) (bool, error) {
+	r := &Allegation{}
+	err := r.Unmarshal(tx.Data)
+	if err != nil {
+		return false, errors.Wrap(action.ErrWrongTxType, err.Error())
+	}
+	err = action.ValidateBasic(tx.RawBytes(), r.Signers(), tx.Signatures)
+	if err != nil {
+		return false, err
+	}
+
+	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
+	if err != nil {
+		return false, gov.ErrGetFeeOptions
+	}
+	err = action.ValidateFee(feeOpt, tx.Fee)
+	if err != nil {
+		return false, err
+	}
+
+	if err := r.ValidatorAddress.Err(); err != nil {
+		return false, err
+	}
+
+	if err := r.MaliciousAddress.Err(); err != nil {
+		return false, err
+	}
+
+	if ctx.EvidenceStore.IsFrozenValidator(r.MaliciousAddress) {
+		return false, action.ErrFrozenValidator
+	}
+
+	if !ctx.EvidenceStore.IsActiveValidator(r.ValidatorAddress) {
+		return false, action.ErrNonActiveValidator
+	}
+
+	if r.ValidatorAddress.Equal(r.MaliciousAddress) {
+		return false, action.ErrInvalidAddress
+	}
+	return true, nil
+}
+
+func (atx allegationTx) ProcessCheck(ctx *action.Context, tx action.RawTx) (ok bool, result action.Response) {
+	ctx.Logger.Debug("Processing 'allegation' transaction for ProcessCheck", tx)
+	ok, result = runAllegationTransaction(ctx, tx)
+	ctx.Logger.Debug("Result 'allegation' transaction for ProcessCheck", ok, result)
+	return
+}
+
+func (atx allegationTx) ProcessDeliver(ctx *action.Context, tx action.RawTx) (ok bool, result action.Response) {
+	ctx.Logger.Debug("Processing 'allegation' transaction for ProcessDeliver", tx)
+	ok, result = runAllegationTransaction(ctx, tx)
+	ctx.Logger.Debug("Result 'allegation' transaction for ProcessDeliver", ok, result)
+	return
+}
+
+func (atx allegationTx) ProcessFee(ctx *action.Context, signedTx action.SignedTx, start action.Gas, size action.Gas) (bool, action.Response) {
+	ctx.Logger.Debug("Processing 'allegation' Transaction for ProcessFee", signedTx)
+	return action.BasicFeeHandling(ctx, signedTx, start, size, 1)
+}
+
+func runAllegationTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	al := &Allegation{}
+	err := al.Unmarshal(tx.Data)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
+	}
+
+	err = ctx.EvidenceStore.PerformAllegation(al.ValidatorAddress, al.MaliciousAddress, al.BlockHeight, al.ProofMsg)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
+	}
+
+	return true, action.Response{Events: action.GetEvent(al.Tags(), "allegation")}
+}
--- a/action/penalization/allegation.go
+++ b/action/penalization/allegation.go
@@ -10,7 +10,6 @@ import (
 	"github.com/Oneledger/protocol/action"
 	"github.com/Oneledger/protocol/action/helpers"
 	"github.com/Oneledger/protocol/data/evidence"
-	gov "github.com/Oneledger/protocol/data/governance"
 	"github.com/Oneledger/protocol/data/keys"
 )
 
@@ -85,7 +84,13 @@ func (atx allegationTx) Validate(ctx *action.Context, tx action.SignedTx) (bool,
 	if err != nil {
 		return false, errors.Wrap(action.ErrWrongTxType, err.Error())
 	}
+
 	err = action.ValidateBasic(tx.RawBytes(), r.Signers(), tx.Signatures)
+	if err != nil {
+		return false, err
+	}
+
+	err = action.ValidateFee(ctx.FeePool.GetOpt(), tx.Fee)
 	if err != nil {
 		return false, err
 	}
@@ -127,16 +132,6 @@ func (atx allegationTx) ProcessFee(ctx *action.Context, signedTx action.SignedTx
 func runAllegationTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
 	al := &Allegation{}
 	err := al.Unmarshal(tx.Data)
-	if err != nil {
-		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrWrongTxType, al.Tags(), err)
-	}
-
-	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
-	if err != nil {
-		return helpers.LogAndReturnFalse(ctx.Logger, gov.ErrGetFeeOptions, al.Tags(), err)
-	}
-
-	err = action.ValidateFee(feeOpt, tx.Fee)
 	if err != nil {
 		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrWrongTxType, al.Tags(), err)
 	}
--- a/action/penalization/allegation.go
+++ b/action/penalization/allegation.go
@@ -17,6 +17,7 @@ import (
 var _ action.Msg = &Allegation{}
 
 type Allegation struct {
+	RequestID        string
 	ValidatorAddress keys.Address
 	MaliciousAddress keys.Address
 	BlockHeight      int64
@@ -65,8 +66,12 @@ func (r Allegation) Tags() kv.Pairs {
 		Key:   []byte("tx.proof"),
 		Value: []byte(r.ProofMsg),
 	}
+	tag6 := kv.Pair{
+		Key:   []byte("tx.requestID"),
+		Value: []byte(r.RequestID),
+	}
 
-	tags = append(tags, tag, tag2, tag3, tag4, tag5)
+	tags = append(tags, tag, tag2, tag3, tag4, tag5, tag6)
 	return tags
 }
 
@@ -132,18 +137,18 @@ func runAllegationTransaction(ctx *action.Context, tx action.RawTx) (bool, actio
 	}
 
 	if ctx.EvidenceStore.IsFrozenValidator(al.MaliciousAddress) {
-		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrFrozenValidator, al.Tags(), err)
+		return helpers.LogAndReturnFalse(ctx.Logger, evidence.ErrFrozenValidator, al.Tags(), err)
 	}
 
 	if !ctx.EvidenceStore.IsActiveValidator(al.ValidatorAddress) {
-		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrNonActiveValidator, al.Tags(), err)
+		return helpers.LogAndReturnFalse(ctx.Logger, evidence.ErrNonActiveValidator, al.Tags(), err)
 	}
 
 	if al.ValidatorAddress.Equal(al.MaliciousAddress) {
 		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrInvalidAddress, al.Tags(), err)
 	}
 
-	err = ctx.EvidenceStore.PerformAllegation(al.ValidatorAddress, al.MaliciousAddress, al.BlockHeight, al.ProofMsg)
+	err = ctx.EvidenceStore.PerformAllegation(al.ValidatorAddress, al.MaliciousAddress, al.RequestID, al.BlockHeight, al.ProofMsg)
 	if err != nil {
 		return helpers.LogAndReturnFalse(ctx.Logger, evidence.ErrCreateAllegationFailed, al.Tags(), err)
 	}
--- a/action/penalization/allegation.go
+++ b/action/penalization/allegation.go
@@ -116,7 +116,12 @@ func (atx allegationTx) ProcessDeliver(ctx *action.Context, tx action.RawTx) (ok
 
 func (atx allegationTx) ProcessFee(ctx *action.Context, signedTx action.SignedTx, start action.Gas, size action.Gas) (bool, action.Response) {
 	ctx.Logger.Debug("Processing 'allegation' Transaction for ProcessFee", signedTx)
-	return action.BasicFeeHandling(ctx, signedTx, start, size, 1)
+	r := &Allegation{}
+	err := r.Unmarshal(signedTx.Data)
+	if err != nil {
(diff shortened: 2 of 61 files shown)


action/errors.go
```diff
@@ -18,6 +18,9 @@ var (
 	ErrNotEnoughFund      = codes.ProtocolError{codes.TxErrInsufficientFunds, "not enough fund"}
 	ErrGasOverflow        = codes.ProtocolError{codes.TxErrGasOverflow, "gas used exceed limit"}
 	ErrInvalidExtTx       = codes.ProtocolError{codes.TxErrInvalidExtTx, "invalid external tx"}
+	ErrFrozenValidator    = codes.ProtocolError{codes.TxErrInvalidExtTx, "frozen validator"}
+	ErrNonFrozenValidator = codes.ProtocolError{codes.TxErrInvalidExtTx, "non frozen validator"}
+	ErrNonActiveValidator = codes.ProtocolError{codes.TxErrInvalidPubKey, "non active validator"}
```
ws4charlie, line 24: These new errors should use proper error codes. TxErrInvalidExtTx and TxErrInvalidPubKey are confusing here.

action/penalization/allegation.go
```diff
@@ -0,0 +1,133 @@
...
+	tag := kv.Pair{
+		Key:   []byte("tx.type"),
+		Value: []byte(r.Type().String()),
+	}
+	tag2 := kv.Pair{
+		Key:   []byte("tx.validator"),
+		Value: r.ValidatorAddress.Bytes(),
+	}
+
+	tags = append(tags, tag, tag2)
```
ws4charlie, line 51: Could you add MaliciousAddress, height and proof to tags? Then explorer will be able to catch full information.

action/penalization/allegation.go
```diff
@@ -0,0 +1,133 @@
...
+func runAllegationTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	al := &Allegation{}
+	err := al.Unmarshal(tx.Data)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
+	}
+
+	err = ctx.EvidenceStore.PerformAllegation(al.ValidatorAddress, al.MaliciousAddress, al.BlockHeight, al.ProofMsg)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
```
ws4charlie, line 129: Can you keep the Log field the same pattern as we did in createProposal.go? The SDK side(not sure which side) need to parse the log string using regular expression or something.

ws4charlie: Update needed in all run****Transaction(...) functions. For Validate, you don't need to change.

action/penalization/release.go
```diff
@@ -0,0 +1,122 @@
...
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
+	}
+
+	blockHeight := ctx.Header.GetHeight()
+	blockCreatedAt := ctx.Header.GetTime()
+
+	options, err := ctx.GovernanceStore.GetEvidenceOptions()
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
```
ws4charlie, line 113: same log pattern

action/penalization/release.go
```diff
@@ -0,0 +1,122 @@
...
+	}
+
+	err = ctx.EvidenceStore.HandleRelease(options, r.ValidatorAddress, blockHeight, blockCreatedAt)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
```
ws4charlie, line 118: same log pattern

action/penalization/release.go
```diff
@@ -0,0 +1,122 @@
...
+
+func runReleaseTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	r := &Release{}
+	err := r.Unmarshal(tx.Data)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
+	}
+
+	blockHeight := ctx.Header.GetHeight()
+	blockCreatedAt := ctx.Header.GetTime()
```
ws4charlie, line 109: Are we using UTC time in DB? Other modules are now using UTC for consistency.

service/tx/service.go
```diff
@@ -182,6 +187,171 @@ func (svc *Service) CreateRawSendPool(args client.SendPoolTxRequest, reply *clie
...
+}
+
+func (svc *Service) Release(args client.ReleaseRequest, reply *client.ReleaseReply) error {
+	validator, err := svc.validators.Get(args.Address)
+	if err != nil {
+		svc.logger.Errorf("validator for address %s not found\n", args.Address)
+		return codes.ErrBadOwner
+	}
+
+	if !svc.evidenceStore.IsFrozenValidator(validator.Address) {
```
ws4charlie, line 313: Have you checked this calling Tx being executed under the same the validator node denoted by args.Address? You can leave the client.ReleaseRequest struct empty and let this Release function to get the node address and fill in penalization.Release{} struct.

action/penalization/allegation.go
```diff
@@ -0,0 +1,133 @@
...
+}
+
+func runAllegationTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	al := &Allegation{}
+	err := al.Unmarshal(tx.Data)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
+	}
+
+	err = ctx.EvidenceStore.PerformAllegation(al.ValidatorAddress, al.MaliciousAddress, al.BlockHeight, al.ProofMsg)
```
ws4charlie, line 127: The allegation ID should be provided by user because tendermint does not guarantee multiple Txs(in same block height) being executed in same order across all nodes. Assume we have 2 allegation Txs(Tx-A, Tx-B) in a block, the execution orders are [Tx-A, Tx-B] and [Tx-B, Tx-A] in Node-0, Node-1 respectively. Then Tx-A, Tx-B in Node-0 will get [ID-1, ID-2], but get [ID-2, ID-1] in Node-1.

service/tx/service.go
```diff
@@ -182,6 +187,171 @@ func (svc *Service) CreateRawSendPool(args client.SendPoolTxRequest, reply *clie
...
+		return codes.ErrSerialization
+	}
+
+	*reply = client.VoteReply{RawTx: packet}
+
+	return nil
+}
+
+func (svc *Service) Allegation(args client.AllegationRequest, reply *client.AllegationReply) error {
+	mv, err := svc.validators.Get(args.MaliciousAddress)
```
ws4charlie, line 244: I saw that the client.AllegationRequest contains an "Address" field. Does that address denote the validator who raises this allegation? Do we allow a validator to raise an allegation under somebody else's node using olclient?

action/penalization/vote.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func runAllegationVoteTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	al := &AllegationVote{}
+	err := al.Unmarshal(tx.Data)
+	if err != nil {
+		return false, action.Response{Log: err.Error()}
+	}
+
+	err = ctx.EvidenceStore.Vote(al.RequestID, al.Address, al.Choice)
```
ws4charlie, line 123: Can you confirm with @pav-kap how to determint if a validator has the right to vote on an allegation.

pav-kap: We are taking a snapshot of validators at the time of the creation of each allegation, right? Those guys can vote

action/penalization/vote.go
```diff
@@ -0,0 +1,129 @@
...
+	}
+	tag2 := kv.Pair{
+		Key:   []byte("tx.voter"),
+		Value: av.Address.Bytes(),
+	}
+	tag3 := kv.Pair{
+		Key:   []byte("tx.choice"),
+		Value: []byte(strconv.Itoa(int(av.Choice))),
+	}
+
```
KevinOneLedger, line 54: Can you include the request ID as well

app/controller.go
```diff
@@ -121,12 +121,26 @@ func (app *App) blockBeginner() blockBeginner {
...
 		app.Context.deliver = storage.NewState(app.Context.chainstate).WithGas(gc)
 
+		err := ManageVotes(&req, &app.Context, app.logger)
+		if err != nil {
+			app.logger.Error("manage votes error", err)
+		}
+
 		// update the validator set
-		err := app.Context.validators.Setup(req, app.Context.node.ValidatorAddress())
+		err = app.Context.validators.Setup(req, app.Context.node.ValidatorAddress(), app.Context.deliver)
```
ws4charlie, line 130: I have a feeling that by explicitly switching to deliver state, you wanted to see the stake/power update immediately in somewhere else(in query or somewhere). If the penalizing happened in last block, are we able to slash the GUILTY validator's stake/power right in last block ender without delay? Delay can makes things more complicated.

cmd/olclient/allegation.go
```diff
@@ -0,0 +1,150 @@
...
+		},
+	}
+
+	packet, err := serialize.GetSerializer(serialize.NETWORK).Serialize(signedTx)
+	if err != nil {
+		ctx.logger.Error("failed to serialize signedTx", err)
+		return err
+	}
+
+	result, err := ctx.clCtx.BroadcastTxCommit(packet)
```
KevinOneLedger, line 142: Can you use "BroadcastTxSync" instead. We recently changed this to prevent timeouts in production.

cmd/olclient/allegation.go
```diff
@@ -0,0 +1,150 @@
...
+	if err != nil {
+		ctx.logger.Error("error in BroadcastTxCommit", err)
+	}
+
+	BroadcastStatus(ctx, result)
```
KevinOneLedger, line 147: For the same reason above you can use:
```

if BroadcastStatusSync(ctx, result) {
	PollTxResult(ctx, result.Hash.String())
}
```

Instead.

cmd/olclient/release.go
```diff
@@ -0,0 +1,141 @@
...
+		},
+	}
+
+	packet, err := serialize.GetSerializer(serialize.NETWORK).Serialize(signedTx)
+	if err != nil {
+		ctx.logger.Error("failed to serialize signedTx", err)
+		return err
+	}
+
+	result, err := ctx.clCtx.BroadcastTxCommit(packet)
```
KevinOneLedger, line 133: Same change as above. Use BroadcastSynch

client/request_types.go
```diff
@@ -54,6 +55,14 @@ type BalanceReply struct {
 	Height int64 `json:"height"`
 }
```
ws4charlie, line 56: If we only allow validator owners using their own node to trigger allegation commands, need to confirm with Pavan, then we don't need the address in below request structs.

cmd/olclient/allegation.go
```diff
@@ -0,0 +1,150 @@
...
+	//Create new Wallet and User Address
+	wallet, err := accounts2.NewWalletKeyStore(keyStorePath)
+	if err != nil {
+		ctx.logger.Error("failed to create secure wallet", err)
+		return err
+	}
+
+	//Verify User Password
+	usrAddress := keys.Address(allegationArgs.Address)
+	authenticated, err := wallet.VerifyPassphrase(usrAddress, allegationArgs.Password)
```
ws4charlie, line 91: Is this usrAddress the stake address? Which address pays for the fee?

cmd/olclient/release.go
```diff
@@ -0,0 +1,141 @@
...
+
+	//Create new Wallet and User Address
+	wallet, err := accounts2.NewWalletKeyStore(keyStorePath)
+	if err != nil {
+		ctx.logger.Error("failed to create secure wallet", err)
+		return err
+	}
+
+	//Verify User Password
+	usrAddress := keys.Address(releaseArgs.Address)
```
ws4charlie, line 81: Same question, which address pays the fee?

cmd/olclient/requests.go
```diff
@@ -0,0 +1,80 @@
...
+)
+
+var requestsCmd = &cobra.Command{
+	Use:   "requests",
+	Short: "Print out vote requests",
+	Run:   GetVoteRequests,
+}
+
+type RequestsArguments struct {
+	address []byte
```
ws4charlie, line 31: By specifying this address, are we able to filter requests by malicious address? Or are we able to get the full list of requests if leaving address as empty?

data/evidence/status.go
```diff
@@ -0,0 +1,108 @@
...
+import (
+	"fmt"
+
+	"github.com/Oneledger/protocol/data/keys"
+	"github.com/Oneledger/protocol/serialize"
+	"github.com/Oneledger/protocol/storage"
+	"github.com/pkg/errors"
+)
+
+type ValidatorStatus struct {
```
ws4charlie, line 12: I'm not sure if we need to keep and maintain this extra active status using 100+ lines of code, because we already have this status in validator store, the lastActive field, and these are maintained by tendermint.

identity/validator_set.go
```diff
@@ -504,11 +524,12 @@ func (vs *ValidatorStore) GetEndBlockUpdate(ctx *ValidatorContext, req types.Req
 				logger.Fatalf("failed  set last purge height, validator: %s", addrHuman)
 			}
 		}
-
 		// delegation
 		ctx.Delegators.UpdateWithdrawReward(height)
+		// vote requests check
+		vs.ExecuteAllegationTracker(ctx, activeCount)
```
ws4charlie, line 530: Just a question. Are we using a dynamic active validator list count to calculate the minimum votes required to pass an allegation?

identity/validator_set_allegation.go
```diff
@@ -0,0 +1,349 @@
...
+			penalizationAmt := new(big.Float).Mul(
+				amt.BigFloat(),
+				big.NewFloat(float64(options.PenaltyBasePercentage)),
+			)
+			penalizationAmt = new(big.Float).Quo(
+				penalizationAmt,
+				big.NewFloat(float64(options.PenaltyBaseDecimals)),
+			)
+			delta := 0.5
+			if penalizationAmt.Sign() < 0 {
```
ws4charlie, line 190: Can penalizationAmt be negative? Are we doing a round here?

status_codes/error.go
```diff
@@ -52,7 +52,9 @@ func WrapError(err error, code int, msg string) *ProtocolError {
 
 */
 var (
-	ErrSerialization = ProtocolError{InternalErrorSerialization, "error in serialization"}
+	ErrSerialization      = ProtocolError{InternalErrorSerialization, "error in serialization"}
+	ErrFrozenValidator    = ProtocolError{InternalErrorSerialization, "error frozen validator"}
```
ws4charlie, line 56: Need to define new error codes for 'ErrFrozenValidator' and 'ErrNonFrozenValidator'

action/penalization/allegation.go
```diff
@@ -0,0 +1,133 @@
...
+	err := r.Unmarshal(tx.Data)
+	if err != nil {
+		return false, errors.Wrap(action.ErrWrongTxType, err.Error())
+	}
+	err = action.ValidateBasic(tx.RawBytes(), r.Signers(), tx.Signatures)
+	if err != nil {
+		return false, err
+	}
+
+	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
```
kingpinXD, line 70: We should access fee options through the context object instead ,ctx.FeePool.GetOpt()

action/penalization/allegation.go
```diff
@@ -0,0 +1,133 @@
...
+
+	if err := r.ValidatorAddress.Err(); err != nil {
+		return false, err
+	}
+
+	if err := r.MaliciousAddress.Err(); err != nil {
+		return false, err
+	}
+
+	if ctx.EvidenceStore.IsFrozenValidator(r.MaliciousAddress) {
```
kingpinXD, line 87: we should move all functions accessing the chainstate to ProcessCheck

action/penalization/release.go
```diff
@@ -0,0 +1,122 @@
...
+	err := r.Unmarshal(tx.Data)
+	if err != nil {
+		return false, errors.Wrap(action.ErrWrongTxType, err.Error())
+	}
+	err = action.ValidateBasic(tx.RawBytes(), r.Signers(), tx.Signatures)
+	if err != nil {
+		return false, err
+	}
+
+	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
```
kingpinXD, line 67: We should access fee options through the context object instead ,ctx.FeePool.GetOpt()

action/penalization/vote.go
```diff
@@ -0,0 +1,129 @@
...
+	err := r.Unmarshal(tx.Data)
+	if err != nil {
+		return false, errors.Wrap(action.ErrWrongTxType, err.Error())
+	}
+	err = action.ValidateBasic(tx.RawBytes(), r.Signers(), tx.Signatures)
+	if err != nil {
+		return false, err
+	}
+
+	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
```
kingpinXD, line 74: We should access fee options through the context object instead ,ctx.FeePool.GetOpt()

service/tx/service.go
ws4charlie: skip

action/penalization/vote.go
ws4charlie: validator list snapshot need to be taken.

action/penalization/vote.go
ws4charlie: confirm with @KevinOneLedger

action/penalization/allegation.go
ws4charlie: Integration test python scripts needed for these Txs.

action/penalization/vote.go
DeV1doR: Note: check if validator is active when voting

client/request_types.go
DeV1doR: VoteRequests address

action/penalization/release.go
DeV1doR: Used this module https://github.com/tendermint/tendermint/blob/master/types/time/time.go and placing with UTC

action/penalization/allegation.go
```diff
@@ -0,0 +1,162 @@
...
+}
+
+func runAllegationTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	al := &Allegation{}
+	err := al.Unmarshal(tx.Data)
+	if err != nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrWrongTxType, al.Tags(), err)
+	}
+
+	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
```
kingpinXD, line 134: I think we should still keep it in validate , because validate is run during the broadcast as well .
Instead we can just replace the access from the gov store by the the context object ctx.FeePool.GetOpt()

action/penalization/release.go
```diff
@@ -0,0 +1,129 @@
...
+}
+
+func runReleaseTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	r := &Release{}
+	err := r.Unmarshal(tx.Data)
+	if err != nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrWrongTxType, r.Tags(), err)
+	}
+
+	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
```
kingpinXD, line 106: Same as allegation.go

action/penalization/vote.go
```diff
@@ -0,0 +1,143 @@
...
+}
+
+func runAllegationVoteTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	al := &AllegationVote{}
+	err := al.Unmarshal(tx.Data)
+	if err != nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrWrongTxType, al.Tags(), err)
+	}
+
+	feeOpt, err := ctx.GovernanceStore.GetFeeOption()
```
kingpinXD, line 119: Same as allegation.go

action/staking/stake.go
```diff
@@ -114,6 +115,10 @@ func (s stakeTx) Validate(ctx *action.Context, tx action.SignedTx) (bool, error)
 	if err != nil {
 		return false, action.ErrNotEnoughFund
 	}
+
```
ws4charlie, line 118: Need to move this state-dependant check to func 'runCheckStake'.

action/staking/unstake.go
```diff
@@ -101,6 +102,9 @@ func (us unstakeTx) Validate(ctx *action.Context, tx action.SignedTx) (bool, err
 		return false, action.ErrInvalidCurrency
 	}
 
+	if ctx.EvidenceStore.IsFrozenValidator(ust.ValidatorAddress) {
```
ws4charlie, line 105: Same as Tx stake. Move state-dependant checks to runCheckUnstake func.

action/staking/withdraw.go
```diff
@@ -104,6 +105,9 @@ func (withdrawTx) Validate(ctx *action.Context, tx action.SignedTx) (bool, error
 		return false, action.ErrInvalidAmount
 	}
 
+	if ctx.EvidenceStore.IsFrozenValidator(draw.ValidatorAddress) {
```
ws4charlie, line 108: Same as stake/unstake Tx.

data/evidence/status.go
```diff
@@ -0,0 +1,108 @@
...
+	}
+	valStatus := &ValidatorStatus{}
+	valStatus, err = valStatus.FromBytes(data)
+	if err != nil {
+		return nil, errors.Wrap(err, "error deserialize validator")
+	}
+	return valStatus, nil
+}
+
+func (es *EvidenceStore) SetValidatorStatus(addr keys.Address, isActive bool, height int64) error {
```
ws4charlie, line 95: So 5000(blocks per day) * 16 * 365 = 3M of validator status records are we going to write into database?

data/evidence/store.go
```diff
@@ -0,0 +1,250 @@
...
+	at.Requests[ID] = true
+
+	err = es.SetAllegationTracker(at)
+	if err != nil {
+		return err
+	}
+	return nil
+}
+
+func (es *EvidenceStore) Vote(requestID string, voteAddress keys.Address, choice int8) error {
```
ws4charlie, line 219: Have you implemented dumping store data to genesis.json?

cmd/olclient/allegation.go
```diff
@@ -0,0 +1,154 @@
...
+	rootPath, err := filepath.Abs(rootArgs.rootDir)
+	if err != nil {
+		return err
+	}
+
+	cfg := &config.Server{}
+
+	//Prompt for password
+	if len(allegationArgs.Password) == 0 {
+		allegationArgs.Password = PromptForPassword()
```
ws4charlie, line 81: @KevinOneLedger @pav-kap Checking password forces validator owner to create a secure wallet account using their private/pub keys before hand. Any problem here?

data/evidence/store.go
DeV1doR: @ws4charlie what do you mean by dumping store?

ws4charlie: > @ws4charlie what do you mean by dumping store?

Yes. Skip this if it's not in requirement doc.

scripts/evidence/test_release.py
```diff
@@ -0,0 +1,65 @@
...
+
+    # checking if validator is frozen
+    fMap = GetFrozenMap()
+    assert valDict['1']['address'] in fMap, 'Validator %s not frozen' % malicious
+
+    print("Test for release successfully passed!")
+
+
+if __name__ == "__main__":
+    test_release()
```
kingpinXD, line 65: Can we add functionality for testing release ? It just tests freezing here

DeV1doR: If it freezing than something with the node. Could you give a more details?

kingpinXD: I mean after it has been frozen , can we also send a release transaction ,in the test .
 - Create a secure wallet account using validator pub/priv key 
 - Send a release request 
 - wait to check if Frozen becomes False

cmd/olclient/allegation.go
kingpinXD: Validator needs the Key Pair for release as well , whatever we decide here should be universal

action/staking/stake.go
ws4charlie: You can either move this check to runCheckStake or just remove it, since 'MinusFromAddress' will return you an error in case of insufficient balance.

cmd/olclient/allegation.go
ws4charlie: Fine.

action/penalization/allegation.go
```diff
@@ -0,0 +1,157 @@
...
+	}
+
+	if !ctx.EvidenceStore.IsActiveValidator(al.ValidatorAddress) {
+		return helpers.LogAndReturnFalse(ctx.Logger, evidence.ErrNonActiveValidator, al.Tags(), err)
+	}
+
+	if al.ValidatorAddress.Equal(al.MaliciousAddress) {
+		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrInvalidAddress, al.Tags(), err)
+	}
+
```
kingpinXD, line 150: We need to check , that we dont already have a allegation request against the malicious validator .

Test case 
Req 1 : Create Alegation  ( Validator 1  -> validator 2)
Req 2 : Create Alegation  ( Validator 1  -> validator 2)
Vote yes : req 1
Vote yes : req 2

Validator 2 gets two deductions in Stake ( Should not happen)

identity/validator_set_allegation.go
```diff
@@ -0,0 +1,344 @@
...
+				}
+				vs.maliciousValidators[addr] = lvh
+			}
+		}
+	}
+	return nil
+}
+
+func (vs *ValidatorStore) ExecuteAllegationTracker(ctx *ValidatorContext, activeCount int64) error {
+	if vs.lastBlockTime == nil {
```
kingpinXD, line 91: Should we allow unstake from a forzen validator ? @pav-kap

scripts/evidence/test_release.py
nm1918: it is covered by tests already, see the code

action/penalization/allegation.go
nm1918: it covered by IsRequestIDBusy function, see the code

kingpinXD: We are talking about two different user IDs here , the same validator cannot have do multiple alegations

nm1918: It was not in functional requirements and never discussed at weekly call, I guess this exception was not realized in code. Please clarify with workflow.

identity/validator_set_allegation.go
kingpinXD: @nm1918 Also here ,we should be checking if if the validator is already frozen , before removing his stake.

scripts/evidence/test_release.py
kingpinXD: What I mean is , follow the following steps 
make reset 
python scripts/evidence/test_release.py
olclient validatorset 

- the last command should show it as Frozen = False , 
For that to happen you need to send a release transaction from the test  , olclient byzantine_fault release

action/penalization/allegation.go
kingpinXD: I think this is an implementation detail ,and not a a functional requirement @pav-kap ?

identity/validator_set_allegation.go
nm1918: you are incorrect, this is another scenario, here the steak was not deducted, here own malicious validator is created because there are not enough votes.
for the scenario where we deducted steak - so there we immediately mark that it is frozen

scripts/evidence/test_release.py
nm1918: Can you clarify, not clear

kingpinXD: you need to add some code , to send a release transaction ,from the frozen validator ,at the end

nm1918: this test related to killing a node and check if it frosted as it not match a vote criteria at N of blocks. For real releasing it better be implemented as a separate test.

identity/validator_set_allegation.go
kingpinXD: I do not fully understand , what I was saying is, 
If two allegations are created against one validator.(two different request ID's)
The staking deduction should happen only once . 
If for some reason you do not want handle it here it can also be handled in action/penalization/allegation.go .
If there is still some confusion please feel free to ping me on slack

DeV1doR: Added for GUILTY and INNOCENT

identity/validator_set_allegation.go
```diff
@@ -214,6 +220,7 @@ func (vs *ValidatorStore) ExecuteAllegationTracker(ctx *ValidatorContext, active
 			} else {
 				logger.Infof("Nothing to withdraw from addr on bounty program: %s\n", bountyAddress.Humanize())
 			}
+			processedValidators[ar.MaliciousAddress.Humanize()] = true
```
kingpinXD, line 223: This still does not solve the problem ,as the processedValidators ,list is created everytime . 
I added a script scripts/evidence.allegation_loadtest.go  , You can use  it to verify .

I think a possible fix would be 
- When you are setting the allegation , you can iterate over existing allegations and see if there is already an allegation existing against the validator.
-  When you freeze the validator, the alllegation must be deleted .

Just a suggestion ,I would be open to any other implementations you think might be better

DeV1doR: If a validator is frozen, we could not create an allegation requests (until defrost process)
If we have a several allegations on the one malicious, it will took into account the first from the requests list, other requests will be removed for this address (will push this part).

PR #632, opened 2020-08-20
Include Penalizing options into governance DB

ws4charlie:


action/governance/finalizeProposal.go
```diff
@@ -298,7 +298,11 @@ func executeConfigUpdate(ctx *action.Context, proposal *governance.Proposal) err
 	}
 	err = ctx.GovernanceStore.WithHeight(ctx.Header.Height).SetStakingOptions(updatedGov.StakingOptions)
 	if err != nil {
-		return errors.Wrap(err, "Setup Fee Options")
+		return errors.Wrap(err, "Setup Staking Options")
+	}
+	err = ctx.GovernanceStore.WithHeight(ctx.Header.Height).SetEvidenceOptions(updatedGov.EvidenceOptions)
```
pav-kap, line 303: Do we need this as part of this or, since this touches governance

cmd/olfullnode/devnet.go
```diff
@@ -543,6 +544,27 @@ func initialState(args *testnetConfig, nodeList []node, option ethchain.ChainDri
 		MaturityTime:            10,
 	}
 
+	// evidence
+	evidenceOption := evidence.Options{
```
pav-kap, line 548: these values will change for kratos, and also based on different environments

action/governance/finalizeProposal.go
DeV1doR: It is just fix for error message

app/application.go
```diff
@@ -110,7 +110,11 @@ func (app *App) setupState(stateBytes []byte) error {
 
 	err = app.Context.govern.SetStakingOptions(initial.Governance.StakingOptions)
 	if err != nil {
-		return errors.Wrap(err, "Setup State")
+		return errors.Wrap(err, "Setup Staking Options")
+	}
+	err = app.Context.govern.SetEvidenceOptions(initial.Governance.EvidenceOptions)
```
kingpinXD, line 115: I think we should add withheight , just to be safe 
app.Context.govern.WithHeight(app.header.Height).SetEvidenceOptions(initial.Governance.EvidenceOptions)

DeV1doR (approved):

app/application.go
DeV1doR: Than for the SetStakingOptions it also must be included, but I did not see an issue, when a self staking part were did.

pav-kap: We have added a few changes for governance options to take height into consideration after self staking is implemented

cmd/olfullnode/devnet.go
```diff
@@ -543,6 +544,27 @@ func initialState(args *testnetConfig, nodeList []node, option ethchain.ChainDri
...
+
+		PenaltyBountyPercentage: 50,
+		PenaltyBountyDecimals:   100,
+
+		PenaltyBurnPercentage: 50,
+		PenaltyBurnDecimals:   100,
+
+		ValidatorReleaseTime: 5,
+
+		AllegationVotesCount: 10,
```
pav-kap, line 563: @DeV1doR do we need this variable?

PR #638, opened 2020-08-27
Upgrade smartcontract abi

kingpinXD:


chains/ethereum/contract/ETH/LockRedeem_Kratos.sol
```diff
@@ -0,0 +1,271 @@
...
+
+    // internal functions
+    function addValidator(address v) internal {
+        validatorList.push(v);
+        validators[v] = uint8(validatorList.length);
+        emit AddValidator(v);
+    }
+
+    function MigrateFromOld() public {
+        require(msg.sender == old_contract);
```
ws4charlie, line 257: Is this expr to make sure address equality?

chains/ethereum/contract/ETH/LockRedeem_Kratos.sol
```diff
@@ -0,0 +1,271 @@
...
+    //Approx
+    // 270 blocks per hour
+    // 5 blocks per min
+    // 28800 old value
+    uint256 LOCK_PERIOD;
+
+    // Variables for migrate
+    address old_contract;
+    uint signaturesrequiredformigration = 0;
+    uint migrationsignaturecount = 0;
```
ws4charlie, line 45: These 3 vars are extra params than existing script?

chains/ethereum/contract/ETH/LockRedeem_Kratos.sol
kingpinXD: yes

chains/ethereum/contract/ETH/LockRedeem_Kratos.sol
kingpinXD: Yes ,these are part of the the Kratos contract ,and not the Kainos one

chains/ethereum/contract/ETH/LockRedeem_Kratos.sol
```diff
@@ -0,0 +1,271 @@
...
+    function addValidator(address v) internal {
+        validatorList.push(v);
+        validators[v] = uint8(validatorList.length);
+        emit AddValidator(v);
+    }
+
+    function MigrateFromOld() public {
+        require(msg.sender == old_contract);
+        migrationsignaturecount = migrationsignaturecount + 1;
+        //addValidator(tx.origin);
```
alexjlan, line 259: why this is removed?

alexjlan: after removing it, how can validator be added to the new contract?

kingpinXD: We are setting up new validators for kratos , so the new Validators will be added throiugh the constructor

PR #660, opened 2020-09-11
Olp 1060 core change

kerhbal:
core protocol changes to support external apps


kerhbal:
will make another one

PR #661, opened 2020-09-11
Olp 1060 core change

kerhbal:
core change to support external apps


app/context.go
```diff
@@ -81,6 +81,10 @@ type context struct {
 	transaction     *transactions.TransactionStore
 	logWriter       io.Writer
 	govupdate       *action.GovernaceUpdateAndValidate
+	extApp          *common.ExtAppData
+	extStores           data.StorageRouter
+	extServiceMap   common.ExtServiceMap
+	controllerFunctions common.Router //External Stores
```
KevinOneLedger, line 87: Try to format this better, also remove the comment //External Stores next to controllerFunctions, it can be confusing.

data/router_test.go
```diff
@@ -3,102 +3,88 @@ package data
...
 const (
-	balanceType = "b"
-	governType  = "g"
-	feeType     = "f"
+	testType = "t"
 )
 
 var (
 	chainstate      *storage.ChainState
-	balanceStore    *balance.Store
```
KevinOneLedger, line 25: The reason why you removed these stores was because they weren't the correct type I'm assuming.

external_apps/common/external_common.go
```diff
@@ -0,0 +1,62 @@
...
+
+func NewExtServiceMap() ExtServiceMap {
+	return ExtServiceMap{}
+}
+
+type ExtAppData struct {
+	// we need txs, data stores, services, block functions
+	ChainState *storage.ChainState
+	ExtTxs []ExtTx
+	ExtStores	map[string]data.ExtStore
```
KevinOneLedger, line 28: I think if you hold alt+ctrl+shift+f it should format your files correctly

external_apps/common/router.go
```diff
@@ -46,4 +46,4 @@ func NewRouter() ControllerRouter {
 	}
 }
 
-var _ Router = &ControllerRouter{}
+var _ Router = &ControllerRouter{}
```
KevinOneLedger, line 49: maybe change the name of this router to something like controllerRouter. It will get confusing with all the routers.

data/router_test.go
kerhbal: yeah, I found this unit test was outdated since we added ExtStore interface to this part, and I just made a simpler unit test to fix this.

external_apps/common/external_common.go
kerhbal: yeah, forgot that

external_apps/common/router.go
kerhbal: yeah sure

app/context.go
kerhbal: yeah sure

KevinOneLedger (approved):

app/controller.go
```diff
@@ -144,14 +145,20 @@ func (app *App) blockBeginner() blockBeginner {
 		//Adds proposals that meet the requirements to either Expired or Finalizing Keys from transaction store
 		//Transaction store is not part of chainstate ,it just maintains a list of proposals from BlockBeginner to BlockEnder .Gets cleared at each Block Ender
 		AddInternalTX(app.Context.proposalMaster, app.Context.node.ValidatorAddress(), app.header.Height, app.Context.transaction, app.logger)
-
-		functionList, err := app.Context.controllerFunctions.Iterate(BlockBeginner)
+		functionList, err := app.Context.extFunctions.Iterate(common.BlockBeginner)
+		functionParam := common.ExtParam{
+			InternalTxStore: app.Context.transaction,
+			Logger:          app.logger,
+			ActionCtx:       *app.Context.Action(&app.header, app.Context.deliver),
```
ws4charlie, line 152: do you need check or deliver state?

kerhbal: @ws4charlie I think for block beginner ender txs we need deliver right, directly commit it.

data/transactions/customTx.go
```diff
@@ -0,0 +1,53 @@
+package transactions
+
```
ws4charlie, line 2: What is this part designed for? no issue, just wanted to know.

kerhbal: it's to add custom prefix for the txs in internal store that waiting to be executed in the block ender.
right now we only have "EXPIRE" and "FINALIZE" for proposals, this part is to add custom one so that they can be treated separately if needed

PR #662, opened 2020-09-15
Olp 1060 all

kerhbal:
bid external app
work flow:
1. A has a domain(or any kind of asset), it's not on sale, but B wants to buy it, B creates a bid conversation and sends a bid offer, every offer sent by B will require B to lock some money to prevent spam.
2. B can cancel this bid conversation as long as it's active.
3. A can agree/reject the deal, or add an counter offer. B's locked money will be released once A made any move except accept the deal.
4. B can agree/reject the deal, or add another bid offer.
...
...
5. If deal, asset ownership will change.
6. The bid conversation comes with a deadline defined by B, if by that time no deal, this whole bid conversation will be automatically expired.


external_apps/bid/bid_action/bidder_decision.go
```diff
@@ -0,0 +1,204 @@
...
+	//5. get the active counter offer
+	activeOffer, err := bidMasterStore.BidOffer.GetActiveOffer(bidderDecision.BidConvId, bid_data.TypeCounterOffer)
+	if err != nil || activeOffer == nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, bid_data.ErrGettingActiveCounterOffer, bidderDecision.Tags(), err)
+	}
+
+	//6. if reject
+	if bidderDecision.Decision != bid_data.RejectBid && bidderDecision.Decision != bid_data.AcceptBid {
+		return helpers.LogAndReturnFalse(ctx.Logger, bid_data.ErrInvalidBidderDecision, bidderDecision.Tags(), err)
+	} else if bidderDecision.Decision == bid_data.RejectBid {
```
kingpinXD, line 121: any particular reason we need the else if? 
an if would be more idiomatic

external_apps/bid/bid_action/create_bid.go
```diff
@@ -0,0 +1,273 @@
...
+	if deadLine.Before(ctx.Header.Time.UTC()) {
+		return "", bid_data.ErrInvalidDeadline
+	}
+
+	//Check if any bid conversation with same asset, owner, bidder already exists in active store
+	store, err := ctx.ExtStores.Get("extBidMaster")
+	if err != nil {
+		return "", bid_data.ErrGettingBidMasterStore.Wrap(err)
+	}
+	bidMasterStore := store.(*bid_data.BidMasterStore)
```
kingpinXD, line 260: You should handle the error ,it might lead to a panic during runtime 
bidMasterStore,ok := store.(*bid_data.BidMasterStore)

kerhbal: yeah I forgot this in action pacakge.

external_apps/bid/bid_action/bidder_decision.go
kerhbal: yeah sure

external_apps/bid/bid_action/counter_offer.go
```diff
@@ -0,0 +1,194 @@
...
+
+	//8. add new counter offer to offer store
+	createCounterOffer := bid_data.NewBidOffer(
+		counterOffer.BidConvId,
+		bid_data.TypeCounterOffer,
+		ctx.Header.Time.UTC().Unix(),
+		counterOffer.Amount,
+		bid_data.CounterOfferAmount,
+	)
+
```
kingpinXD, line 152: Do we need to make make sure the conversation does not expire in this exact block ?

kerhbal: yeah there is in line 110

kerhbal: ok you are talking about expire in exact this block.
then I think we are fine, because in next block it will be expired

kingpinXD: Yeah , but the bidder wont have a chance to agree or disagree to it ?

kerhbal: yeah, but if we allow one or two blocks for the other party to respond, then the deadline showing in bidConv will be no longer accurate.

kingpinXD: I was thinking ,maybe we can increment the expiry everytime a counter offer is made .by a fixed amount .

kerhbal: yeah good idea, or maybe we can just increment the expiry after any new changes

external_apps/bid/bid_action/owner_decision.go
```diff
@@ -0,0 +1,200 @@
...
+	}
+
+	//8. change offer status to inactive and add it back to bid offer store
+	err = DeactivateOffer(true, bidConv.Bidder, ctx, activeOffer, bidMasterStore)
+	if err != nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, bid_data.ErrDeactivateOffer, ownerDecision.Tags(), err)
+	}
+
+	//9. close the bid conversation
+	err = CloseBidConv(bidConv, bidMasterStore, bid_data.BidStateSucceed)
```
kingpinXD, line 152: If we have multiple conv for one asset , The other conversations would just expire eventually ?

kingpinXD (approved):

KevinOneLedger (commented):
Looks good

KevinOneLedger (approved):

external_apps/bid/bid_action/create_bid.go
```diff
@@ -0,0 +1,273 @@
...
+	return runCreateBid(ctx, tx)
+}
+
+func (c CreateBidTx) ProcessFee(ctx *action.Context, signedTx action.SignedTx, start action.Gas, size action.Gas) (bool, action.Response) {
+	return action.BasicFeeHandling(ctx, signedTx, start, size, 1)
+}
+
+func runCreateBid(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	// if this is to create bid conversation, everything except bidConvId is needed
+	// if this is just to add an offer from bidder, only needs bidConvId, bidder(to sign), amount
```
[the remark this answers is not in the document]
kerhbal: there are two scenarios:
1. create bid conversation AND bid offer(create both in one tx)
2. add an offer to an existing bid conversation.

the reason for this is that if we separate create bid conversation and create bid offer into two separate tx, we always need to do 2 txs at beginning, because an empty bid conversation without any offers makes no sense

external_apps/bid/bid_action/bidder_decision.go
```diff
@@ -0,0 +1,205 @@
...
+		return helpers.LogAndReturnFalse(ctx.Logger, bid_data.ErrWrongBidder, bidderDecision.Tags(), err)
+	}
+
+	//5. get the active counter offer
+	activeOffer, err := bidMasterStore.BidOffer.GetActiveOffer(bidderDecision.BidConvId, bid_data.TypeCounterOffer)
+	if err != nil || activeOffer == nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, bid_data.ErrGettingActiveCounterOffer, bidderDecision.Tags(), err)
+	}
+
+	//6. if reject
```
ws4charlie, line 118: I have a feeling that the Offer Acrive Status and bidder_decision Tx made this bid app more complicated than it's supposed to. But it's still a good app for tutorial purpose.

kerhbal: because we need to know which offer is the active one and to respond to that. the logic here is any offer/counter offer without being replied is active one as long as this whole bid conversation is ongoing.

external_apps/bid/bid_block_func/bid_block_func.go
```diff
@@ -0,0 +1,137 @@
...
+	"github.com/Oneledger/protocol/external_apps/bid/bid_action"
+	"github.com/Oneledger/protocol/external_apps/bid/bid_data"
+	"github.com/Oneledger/protocol/external_apps/common"
+	"github.com/google/uuid"
+	abci "github.com/tendermint/tendermint/abci/types"
+	"time"
+)
+
+// Function for block Beginner
+func AddExpireBidTxToQueue(i interface{}) {
```
ws4charlie, line 15: We queued the expired Txs at a block beginner and expire them in the same block's ender?

kerhbal: yeah

external_apps/bid/bid_action/owner_decision.go
kerhbal: and actually if we do not reject those other offers we are still fine.
1. they will be expired
2. if we make counter offers to two people(by design this is not recommended), the first one that deals will get the asset. and the second one will fail to agree to the deal, because the asset right now is no longer valid, since the ownership has changed, and if will fail at asset validation part.

ws4charlie (approved):

external_apps/bid/bid_action/bidder_decision.go
ws4charlie: Yeah. This is where this app complicated in. We can go with this solution.

PR #670, opened 2020-10-07
Olp 1159-undelegate

kerhbal:
OLP-1159: undelegate
OLP-1166: query the total amount undelegated by a Delegator, this one will return pending amount, matured amount for one address
OLP-1169: query total amount contributed to delegation pool, this one will return active amount, pending amount, matured amount as total from all the addresses

data store:
added some wrappers with unit tests

integration test:
same delegator undelegates in same block: the undelegation amounts will become one entry in pending store with the sum of undelegated amounts
same delegator undelegates in different block: the undelegation amounts will become different entries in pending store for each height
malicious undelegate


app/application.go
```diff
@@ -169,6 +169,11 @@ func (app *App) setupState(stateBytes []byte) error {
 	}
 	app.Context.feePool.SetupOpt(&initial.Governance.FeeOption)
 
+	err = app.Context.govern.WithHeight(app.header.Height).SetNetworkDelegOptions(initial.Governance.DelegOptions)
```
kingpinXD, line 172: I think we should use ,hard coded constants to set options for now . Until we decide to add things to genesis

data/network_delegation/init.go
```diff
@@ -7,11 +7,17 @@ import (
...
-
 var logger *log.Logger
 var prefixMap map[DelegationPrefixType]string
 
+const (
+	MatureType  DelegationPrefixType = 0x103
+	PendingType DelegationPrefixType = 0x102
+	ActiveType  DelegationPrefixType = 0x101
+
+	DELEGATION_POOL_KEY = "00000000000000000001"
```
kingpinXD, line 18: We can probably put the hardcoded values of the delegation options from here, and call this from application.do

kingpinXD (commented):
Great Work , Minor Comments please have a look

kingpinXD (approved):

PR #673, opened 2020-10-12
Minor fix

kingpinXD:
The problem 
Multiple allegations should not be created against the same validator . We have to check if there is already an allegation against the validator ,and not create a new one . 
- This should work if there are multiples requests in the same block as well . 

@ws4charlie @KevinOneLedger @DeV1doR  , Have a look at his solution , let me know if I missed some edge cases 
f72b26d91ae026f3bedc35cde8f5698eb4feb4e8 , this commit is also relevant to verify this . (This pr contains those changes as well, I mentioned the commit for easy verification )

This doc is for reference of the testing done 
https://docs.google.com/document/d/1a_9iN-v6FbAusgDKpa_bhOr3mZjxd-rLX4FCcWKTt1A/edit


kingpinXD:
Staking
# Cannot freeze frozen Validator
# Frozen Validator cannot stake,unstake ,withdraw stake  -> error frozen validator
# Validator  frozen = true  block X , Stake and Active = false in block X +1

 Rewards
# Frozen validator cannot accumulate rewards  , checked rewards diff between frozen and active
# Frozen
# {
#     "address": "0lt05e5e4ea96d1a6fd36acdd7196fe738f38191c06",
#     "matureBalance": "84383561643835616437",
#     "pendingAmount": "0",
#     "totalAmount": "84383561643835616437",
#     "withdrawnAmount": "0"
# }
 Active
# {
#     "address": "0lt096a4dee464e74c8f0b667835ba04413998fd958",
#     "matureBalance": "128493150684931506835",
#     "pendingAmount": "9589041095890410958",
#     "totalAmount": "138082191780821917793",
#     "withdrawnAmount": "0"
# }
# Frozen Validator can withdraw block rewards

 Penalties
# Release tx does not refund stake
# Released Validator can be penalized again

 Governance
# Frozen Validator can vote for proposals created before it is frozen
# Frozen Validators cannot vote for for proposals created after it is frozen

data/evidence/allegation.go
```diff
@@ -45,6 +46,32 @@ type AllegationTracker struct {
...
+	if err != nil {
+		return
+	}
+	requestIdList := make([]string, len(at.Requests))
+	for k := range at.Requests {
+		requestIdList = append(requestIdList, k)
+	}
+	sort.Strings(requestIdList)
+	countMap := make(map[string]bool)
+	for _, r := range requestIdList {
```
KevinOneLedger, line 60: Can you explain this loop? I see that you are deleting request ID's from your allegation tracker if they exist in the store. If they don't exist in the store then you save it to a count map using its address. If the request has an address already in the count map you delete it from the store. Is the allegation request already in the store at this point?

kingpinXD: yes, Thats exactly what its for . 
The motive is to clean , two requests with the same malicious address so that the the tracker at any point of time has oa max 1 request id per valdiator

data/evidence/allegation.go
```diff
@@ -45,6 +46,32 @@ type AllegationTracker struct {
...
+		ar, err := es.GetAllegationRequest(r)
+		if err != nil {
+			delete(at.Requests, r)
```
ws4charlie, line 63: What scenario is this delete for?

kingpinXD: This is a fail-safe , in case we have a request-Id , which does not have an associated allegation request

ws4charlie (dismissed):

Makefile
```diff
@@ -49,13 +49,15 @@ utest:
...
 		github.com/Oneledger/protocol/data/network_delegation \
+		github.com/Oneledger/protocol/data/evidence \
 		github.com/Oneledger/protocol/action/transfer \
 		github.com/Oneledger/protocol/serialize \
 		github.com/Oneledger/protocol/utils \
 		github.com/Oneledger/protocol/rpc \
 		github.com/Oneledger/protocol/identity \
 		github.com/Oneledger/protocol/app \
 		github.com/Oneledger/protocol/action/staking \
+		github.com/Oneledger/protocol/action/penalization \
```
ws4charlie, line 60: are we able to use a same package name, 'evidence'？

action/penalization/vote.go
```diff
@@ -0,0 +1,137 @@
...
+}
+
+func runAllegationVoteTransaction(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	al := &AllegationVote{}
+	err := al.Unmarshal(tx.Data)
+	if err != nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrWrongTxType, al.Tags(), err)
+	}
+
+	if ctx.EvidenceStore.IsFrozenValidator(al.Address) {
```
ws4charlie, line 123: I think below 'IsActiveValidator' check already covered this 'Frozen' case, but it's still good to double check it.

action/staking/unstake.go
```diff
@@ -126,11 +127,19 @@ func runCheckUnstake(ctx *action.Context, tx action.RawTx) (bool, action.Respons
...
+	if ctx.EvidenceStore.IsFrozenValidator(ust.ValidatorAddress) {
+		return false, action.Response{Log: evidence.ErrFrozenValidator.Error()}
+	}
+
 	unstake := identity.Unstake{
 		Address: ust.ValidatorAddress,
 		Amount:  ust.Stake.Value,
 	}
 
+	if ctx.EvidenceStore.CheckRequestExists(ust.ValidatorAddress) {
```
ws4charlie, line 139: Is this the edge case that we disallow unstake when validator is under an on-going allegation?

app/context.go
```diff
@@ -178,6 +182,7 @@ func newContext(logWriter io.Writer, cfg config.Server, nodeCtx *node.Context) (
 	_ = action_gov.EnableGovernance(ctx.actionRouter)
 	_ = action_gov.EnableInternalGovernance(ctx.internalRouter)
 	_ = staking.EnableStaking(ctx.actionRouter)
+	_ = action_pen.EnablePenalization(ctx.actionRouter)
```
ws4charlie, line 185: Are we able to unify package name

data/evidence/store.go
```diff
@@ -0,0 +1,273 @@
...
+		prefixKey,
+		storage.Rangefix(string(prefixKey)),
+		true,
+		func(key, value []byte) bool {
+			lvh, err := (&LastValidatorHistory{}).FromBytes(value)
+			if err != nil {
+				fmt.Println("failed to deserialize susp validator")
+				return false
+			}
+			if !lvh.IsFrozen() {
```
ws4charlie, line 103: Iterating thru 'Suspicious' validators means iterating thru 'Frozen' validators?

data/evidence/store.go
```diff
@@ -0,0 +1,273 @@
...
+		vote := ar.Votes[i]
+		if vote.Address.Equal(voteAddress) {
+			return fmt.Errorf("You have been already voted on this request")
+		}
+	}
+	vote := &AllegationVote{
+		Address: voteAddress,
+		Choice:  choice,
+	}
+	ar.Votes = append(ar.Votes, vote)
```
ws4charlie, line 266: Can we sort this Votes list?

data/evidence/store.go
```diff
@@ -0,0 +1,273 @@
...
+	}
+
+	at, err := es.GetAllegationTracker()
+	if err != nil {
+		return err
+	}
+
+	at.Requests[ID] = true
+
+	err = es.SetAllegationTracker(at)
```
ws4charlie, line 235: Can we sort the map inside 'at' before save?

data/evidence/store.go
kingpinXD: I think this was named so because of the key thats used in the iterator function .

action/penalization/vote.go
kingpinXD: Yeah ,  I just left it how it was .

Makefile
kingpinXD: Good point ,I think penalization makes more sense ,I can rename the data layer package

data/evidence/store.go
kingpinXD: Hmm ,the map is being cleaned at every block 
CleanTracker() , does exactly that . 

Do you mean to call the same function here as well , for additional safety ?

data/evidence/vote.go
```diff
@@ -0,0 +1,117 @@
...
+					delete(cv.Addresses, hAddr)
+				}
+			}
+		}
+	}
+	for _, address := range vb.Addresses {
+		hAddr := address.String()
+		cv.Addresses[hAddr]++
+	}
+	dat, err := serialize.GetSerializer(serialize.PERSISTENT).Serialize(cv)
```
ws4charlie, line 112: Can we sort the map as well?

Makefile
ws4charlie: > Good point ,I think penalization makes more sense ,I can rename the data layer package

Thanks. Take your time.

action/staking/unstake.go
kingpinXD: yes

data/evidence/vote.go
kingpinXD: It gets stores as a map again , so even if I sort the keys ,ill need to add it back to the map .
Do you mean to convert it to an ordered map ?

Makefile
kingpinXD: actually renamed the action package to evidence ,since we already have the options names as evidenceoptions in kratos, it would be more uniform this way

ws4charlie (dismissed):

ws4charlie (approved):

ws4charlie (approved):

PR #684, opened 2020-10-26
Develop feature/olp 1230

ws4charlie:
1. loadtest framework utilizing existing integration test functions
2. 'make loadtest' to run loadtest locally
3. In devnet, use 'python scripts/loadtest/run_tests.py' to start loadtest on fullnode.
3. Load test for Delegate Tx.


scripts/network_delegation/addNetworkDelegation_load.py
```diff
@@ -0,0 +1,33 @@
+from sdk import *
+
+cfg_dev = TestConfig(fullnode_dev, 11000, 100, INTERVAL_NORMAL)
```
kingpinXD, line 3: I think eventually we should add a function to create the config ,by randomly picking nodes ,based on how many threads the user wants to run

kingpinXD (approved):

PR #688, opened 2020-10-30
Olp 1267 remove delegation withdraw and matured amount

kerhbal:
1. remove delegation withdraw
2. remove concept of matured amount
3. directly mature the amount to delegator's balance
4. change ToCoinWithBase to ToCoin since sdk will pass the number that already has 18 digit base
5. test

### DIFF
--- a/app/controller.go
+++ b/app/controller.go
@@ -3,7 +3,6 @@ package app
 import (
 	"encoding/hex"
 	"fmt"
-
 	"math"
 	"math/big"
 	"runtime/debug"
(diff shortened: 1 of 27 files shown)


app/controller.go
```diff
@@ -771,3 +766,36 @@ func ManageVotes(req *RequestBeginBlock, ctx *context, logger *log.Logger) error
 	}
 	return nil
 }
+
+func addMaturedAmountsToBalance(ctx *context, logger *log.Logger, req *RequestBeginBlock) {
```
ws4charlie, line 770: Is it a good idea to move this func to delegation store? We already have lots of funcs like this in controller.go.

kerhbal: this involves both delegation store and balance store, if we move it into delegation store that means we are accessing balance store inside delegation store, is that ok?

ws4charlie (dismissed):

ws4charlie (approved):

PR #689, opened 2020-10-30
Develop feature/olp 1252

ws4charlie:
1. Implemented 'Reinvest Rewards' transaction.
2. Integration test for ReinvestRewards transaction.
3. Defined pool names as constants and unified the way we get pool address by its name


action/network_delegation/reinvest_rewards.go
```diff
@@ -0,0 +1,148 @@
...
+	//Add Delegation
+	//Get Delegation Pool
+	poolList, err := ctx.GovernanceStore.GetPoolList()
+	if err != nil {
+		return helpers.LogAndReturnFalse(ctx.Logger, gov.ErrPoolList, invest.Tags(), err)
+	}
+	if _, ok := poolList["DelegationPool"]; !ok {
+		return helpers.LogAndReturnFalse(ctx.Logger, action.ErrPoolDoesNotExist, invest.Tags(), err)
+	}
+	delagationPool := poolList["DelegationPool"]
```
kingpinXD, line 130: We should probably add the strings as constants , probably in data/governanace/store.go

service/tx/network_delegation.go
```diff
@@ -109,6 +109,38 @@ func (s *Service) WithdrawDelegRewards(args client.WithdrawDelegRewardsRequest,
...
+	}
+
+	uuidNew, _ := uuid.NewUUID()
+	feeAmount := s.feeOpt.MinFee()
+	tx := &action.RawTx{
+		Type: action.REWARDS_REINVEST_NETWORK_DELEGATE,
+		Data: data,
+		Fee: action.Fee{
+			Price: action.Amount{Currency: "OLT", Value: *feeAmount.Amount},
+			Gas:   80000,
```
kingpinXD, line 130: Just curious ,does the trasaction fail with not enough gas ,for values less than 80000  ?

ws4charlie: > Just curious ,does the trasaction fail with not enough gas ,for values less than 80000 ?

I don't know the exact number of min gas that cause a failure, 80000 is an estimated gas big enough to guarantee a success.

kerhbal: I failed at 40000 for some txs before

action/network_delegation/reinvest_rewards.go
ws4charlie: > We should probably add the strings as constants , possibly in data/governanace/store.go

Good idea to unify these.

kingpinXD (dismissed):

kingpinXD (approved):

kingpinXD (approved):

kingpinXD (approved):

PR #698, opened 2020-11-11
OLP-1269: Setup docker compose

KevinOneLedger:
1. Created default docker-compose files for build and node containers
2. Added scripts for starting, updating and stopping the network (Written in bash, might need to change this for windows)

### DIFF
--- a/DOCKER-network/docker-compose.yml
+++ b/DOCKER-network/docker-compose.yml
@@ -3,7 +3,7 @@ version: "3"
 services:
   0-Node:
     container_name: 0-Node
-    image: appbuilder
+    image: olprotocol
     ports:
       - "26602:26602"
     volumes:
@@ -20,7 +20,7 @@ services:
 
   1-Node:
     container_name: 1-Node
-    image: appbuilder
+    image: olprotocol
     volumes:
       - ~/oldata/docker/bin:/home/ubuntu/go/bin
       - ~/oldata/docker/1-Node:/opt/data/devnet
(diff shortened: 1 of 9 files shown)


DOCKER-network/docker-compose.yml
```diff
@@ -0,0 +1,87 @@
...
+    image: olprotocol
+    volumes:
+      - ~/oldata/docker/bin:/home/ubuntu/go/bin
+      - ~/oldata/docker/1-Node:/opt/data/devnet
+    command: >
+      bash -c "cd /opt/data/devnet &&
+              olfullnode node --root /opt/data/devnet >> /opt/data/devnet/olfullnode.log 2>&1"
+    networks:
+      vpcbr:
+        ipv4_address: 10.5.0.11
```
Jedi2002, line 32: Is there a reason for not using the bridge network ?

DOCKER-network/stop.sh
```diff
@@ -0,0 +1,5 @@
+#!/bin/bash
+
+# shellcheck disable=SC2046
+docker container kill $(docker ps -q)
+docker system prune -f
```
Jedi2002, line 5: I think we should remove system prune , and just remove this network instead .

kingpinXD, line 5: Can we just remove this specific network instead of using system prune

kingpinXD (approved):
Minor comment . Looks fine otherwise

DOCKER-network/stop.sh
KevinOneLedger: Yea I can look into that, I was assuming that developers would run one network on their machine at a time.

PR #703, opened 2020-11-16
OLP-1289: Delegator Rewards not maturing

KevinOneLedger:
- Moved maturity logic outside of handle delegation rewards.
- Delegation pool can be zero to mature rewards for delegators.


kerhbal (commented):
since `matureDelegationRewards` is moved outside from `handleDelegationRewards`, I think u can remove `req *RequestBeginBlock, logger *log.Logger` in the `handleDelegationRewards`'s signature

kerhbal (approved):

PR #711, opened 2020-11-26
Olp 1318 Wrapper for cli command to be used in both native and docker development

kerhbal:
- command wrapper, like turn `olclient list` into `python $OLROOT/protocol/DOCKER-network/dockernet run 0-Node "cd /opt/data/devnet && olclient list"`
- node folder wrapper
  - for external access, like from python script, we will be able to get the correct node folder depends on native or docker
  - for internal access from docker container, automatically change keystore path to `./keystore` as we are always in `/opt/data/devnet` when using command wrapper
- keystore population
  - automatically populate keystore to all other nodes when account is created, because we are not always using same node for creating account and signing tx, when they are different, docker container cannot get other node's keystore file
- added some default parameters in `devnet.go` and `dockernet`


scripts/getValidators_docker
```diff
@@ -0,0 +1,15 @@
+#!/bin/bash
+# find node folder from docker inspect result
+node_folder=$(docker inspect 0-Node | sed -n 's/^.*"Source"\: "\(.*\)\/[0-9]-Node.*$/\1/p')
```
KevinOneLedger, line 3: Can we get this from an environment variable instead?
This is probably more accurate but, just thinking to simplify it. Not really and issue.

scripts/status_docker
```diff
@@ -0,0 +1,9 @@
+#!/bin/bash
+
+# List of pids if the chain is running
+# No output if the chain isn't running
+
+echo -n "all running nodes: "
+echo
+
+docker ps | grep '\d\-Node'
```
KevinOneLedger, line 9: you can use the filter:
docker ps -f network=docker-network_vpcbr 
To get running nodes

scripts/testsend_docker
```diff
@@ -0,0 +1,14 @@
+#!/bin/bash
+# find node folder from docker inspect result
+node_folder=$(docker inspect 0-Node | sed -n 's/^.*"Source"\: "\(.*\)\/[0-9]-Node.*$/\1/p')
```
KevinOneLedger, line 3: I mentioned above the node_folder can be a env variable.

KevinOneLedger (approved):

scripts/status_docker
kerhbal: as mentioned in another pr, my network name is different, not sure why

scripts/getValidators_docker
kerhbal: yeah, agree, this will be much simpler

KevinOneLedger (approved):

PR #713, opened 2020-12-01
Develop feature/olp 1338

ws4charlie:
Missed writing pending unstaked amount into DB on chain initialization.


KevinOneLedger (approved):

kerhbal (approved):

scripts/cleanNode
```diff
@@ -0,0 +1,32 @@
+#!/bin/bash
```
kingpinXD, line 1: Is this script a copy of gov/clean ?

kingpinXD (approved):

scripts/cleanNode
kingpinXD: Oh okay ,this one does not start the nodes .

PR #719, opened 2020-12-04
Feature/add witness

kingpinXD:


KevinOneLedger (commented):
You also need to dump the witnesses to the genesis in the save_state cmd

kingpinXD:
> You also need to dump the witnesses to the genesis in the save_state cmd

Yeah , I just made this change so that develop is testable right now , the dump would be a bit more involved and require more time to be merged .

PR #724, opened 2021-02-18
OLP-1426: Saved Witnesses as separate section in genesis file

KevinOneLedger:
1. Updated unit tests
2. Refactored function in witness store
3. Dumped the Witnesses separately to the genesis
4. Tested by starting network, 
  - deploying smart contract
  - stopping
  - saving state
  - starting back up
  - performed lock and redeem

### DIFF
--- a/cmd/olfullnode/devnet.go
+++ b/cmd/olfullnode/devnet.go
@@ -578,6 +578,7 @@ func initialState(args *testnetConfig, nodeList []node, option ethchain.ChainDri
 	}
 	balances := make([]consensus.BalanceState, 0, len(nodeList))
 	staking := make([]consensus.Stake, 0, len(nodeList))
+	witnesses := make([]identity.Witness, 0, len(nodeList))
 	penalities := make([]identity.Penalty, 0)
 	rewards := rewards.RewardMasterState{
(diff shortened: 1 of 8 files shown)


cmd/olfullnode/devnet.go
```diff
@@ -667,7 +668,14 @@ func initialState(args *testnetConfig, nodeList []node, option ethchain.ChainDri
 			Name:             node.validator.Name,
 			Amount:           *balance.NewAmountFromInt(node.validator.Power),
 		}
+		wt := identity.Witness{
+			Address:     node.validator.Address.Bytes(),
+			PubKey:      pubkey,
```
kingpinXD, line 673: Does a witness need this ? or just the ECDSA key is enough ?

KevinOneLedger: I'm not sure if the other key is really used, I just wanted to do the refactor to use an actual witness struct. The implementation of the Witness itself is something I didn't want to modify.

kingpinXD: makes sense .

kingpinXD (dismissed):
lgtm

freddyli7 (approved):

PR #729, opened 2021-03-25
Evm intergration

DeV1doR:


action/evm.go
```diff
@@ -0,0 +1,231 @@
...
+// 	// NOTE: if the validator set hash is missing the hash will be returned as nil,
+// 	// so we need to check for this case to prevent a panic when calling Bytes()
+// 	if tmBlockHash == nil {
+// 		return ethcmn.Hash{}
+// 	}
+
+// 	return ethcmn.BytesToHash(tmBlockHash.Bytes())
+// }
+
+// GetHashFn implements vm.GetHashFunc for protocol. It handles 3 cases:
```
alexjlan, line 71: 3 cases? but you just show 2.

action/smart_contract/execute.go
```diff
@@ -0,0 +1,166 @@
...
+	}
+	return true, nil
+}
+
+func (s scExecuteTx) ProcessCheck(ctx *action.Context, tx action.RawTx) (bool, action.Response) {
+	ctx.Logger.Detail("Processing SC Deploy Transaction for CheckTx", tx)
+	// NOTE: We do not need to run a logic, but only calculate a gas
+	// maybe it will be required a stateDB copy without deliver and clear it's dirties
+	// TODO: Add gas calculation here
+	return true, action.Response{GasUsed: -1}
```
alexjlan, line 102: why not directly run the tx, the checkstate will be dropped as you expected.

service/query/evm.go
```diff
@@ -0,0 +1,239 @@
...
+
+// EVMCall call smart contract code to get the result
+func (svc *Service) EVMCall(args client.SendTxRequest, reply *client.EVMCallReply) error {
+	height := svc.contracts.State.Version()
+	stateDB := action.NewCommitStateDB(svc.contracts, svc.accountKeeper, svc.logger)
+	bhash := stateDB.GetHeightHash(uint64(height))
+	stateDB.SetBlockHash(bhash)
+	// TODO: Change this
+	header := &abci.Header{
+		ChainID: "test-1",
```
alexjlan, line 132: chain id why hard coded?

service/query/evm.go
```diff
@@ -0,0 +1,239 @@
...
+
+// EVMEstimateGas call smart contract code to get the gas
+func (svc *Service) EVMEstimateGas(args client.SendTxRequest, reply *client.EVMEstimateGasReply) error {
+	height := svc.contracts.State.Version()
+	stateDB := action.NewCommitStateDB(svc.contracts, svc.accountKeeper, svc.logger)
+	bhash := stateDB.GetHeightHash(uint64(height))
+	stateDB.SetBlockHash(bhash)
+	// TODO: Change this
+	header := &abci.Header{
+		ChainID: "test-1",
```
alexjlan, line 200: same here

service/tx/service.go
```diff
@@ -65,16 +67,39 @@ func NewService(
...
 // a signed transaction
 // TODO: deprecate this
 func (svc *Service) SendTx(args client.SendTxRequest, reply *client.CreateTxReply) error {
-	send := transfer.Send{
-		From:   keys.Address(args.From),
-		To:     keys.Address(args.To),
-		Amount: args.Amount,
-	}
-	data, err := send.Marshal()
+	data, t, err := getSendTxContext(args)
```
alexjlan, line 102: the SendTx rpc was just simple send, maybe should make a separated rpc for contract?

action/evm.go
DeV1doR: Not removed a comment from the etheremint. Will remove as we do not need to make a same logic as there.

action/smart_contract/execute.go
DeV1doR: Because it will run a evm code which is not a part of the txChecker. Also it breaks a state for the stateDB and we have an API to calculate the aprox gas.

DeV1doR: Also by the tendermint what I found that we do not need a full logic during check

service/tx/service.go
DeV1doR: Made it similar to the ethereum logic.

action/smart_contract/execute.go
alexjlan: for tendermint, it doesn't require checktx to execute full logic, but as a blockchain, the checktx result different from delivery tx might cause apphash difference in certain cases.

DeV1doR: @alexjlan https://docs.tendermint.com/master/spec/abci/abci.html#checktx
```
CheckTx need not execute the transaction in full, but rather a light-weight yet stateful validation, like checking signatures and account balances, but not running code in a virtual machine.
```
From the code above it is deterministic, but in the case of VM the statedb needs to be encapsulated as the bunch of logic are executed at the end (blockEnder). If it will be running there, it may cause the apphash difference as:
```
The transaction may come from an external user or another node
```
as some state is not already on the node, which will execute it.

I can add more checks related to the state, but rather than run a full evm execution. Also it will slow down the network in case of high gas limit contracts.

service/query/evm.go
DeV1doR: will fix, not a critical point for the simulation of vm.

service/tx/service.go
DeV1doR: I also thought about to make it separatelly, but anyway it is more convenient to use in the sendTransaction.

action/smart_contract/execute.go
DeV1doR: Added

alexjlan (approved):

PR #740, opened 2021-05-10
Add fix for balances, revert back to skip validation on process check

DeV1doR:


app/controller.go
```diff
@@ -268,8 +263,6 @@ func (app *App) txChecker() txChecker {
 			Codespace: "",
 		}
 
-		app.resetInternalState(app.Context.check)
```
DeV1doR, line 271: This could clear the whole journal for the blockEnder which may cause a hash id error.

alexjlan (approved):

alexjlan:
if the check tx is failing because of executing the evm tx, it means there might be something wrong in the implementation. approved this pr for urgency request.

